PR #1, opened 2026-04-11
Add DraftPane, PR summary generation, and fix GitHub PR fetch

SDSLeon:
- Add `DraftPane` component enabling draft thread views as first-class panes in the multi-pane layout alongside active threads
- Refactor pane drag-drop data to use generic `paneId` instead of `threadId`, unblocking drag support for non-thread panes
- Add `generatePrSummary` IPC channel wired through main/preload for AI-powered PR description generation
- Fix GitHub PR creation to fetch PR metadata via `gh pr view` after create rather than using the broken inline `--json` flag


SDSLeon:
isolated validation

PR #21, opened 2026-05-24
Aggregate PR checks status in the review overlay

SDSLeon:
- Feature + refactor: unify PR-level and cached PR-detail checks state so the review overlay reflects the latest checks result instead of stale partial data.
- Refresh pending PR data when checks are still unresolved, including cached PR details, to keep the overlay in sync with the remote PR.
- Surface checks status in the tab pill so the Checks icon shows success, warning, or failure at a glance.
- Add coverage for the refresh path and the tab icon color states.


SDSLeon:
hi github!

PR #273, opened 2026-07-09
Add computer-use MCP for desktop control across providers

SDSLeon:
## Summary
Revives **PR #142** ("Add computer-use MCP for desktop control across providers"), which had drifted ~193 commits behind `next`, ports it onto the current runtime, live-probes it on a real Windows 11 host, and hardens it through a 5-dimension adversarial review + two executor rounds. Result: an **opt-in, per-thread** capability letting an agent see and drive native desktop apps (screenshots, window enumeration, mouse/keyboard/scroll/drag, launch) over a **localhost-only** MCP server, wired uniformly across every provider.

Three commits: baseline port → correctness/security hardening → performance.

## Architecture
- **Main process** `src/main/computer-use/`: `ComputerUseMcpIngress` (localhost HTTP MCP, binds `127.0.0.1` only) + platform drivers `windows.ts` (persistent PowerShell 5.1 + inline C#/P-Invoke host) and `macos.ts` (osascript/JXA). Shared transport `src/main/mcp/StreamableHttpMcpIngress.ts`.
- **Supervisor**: `computerUseMcp/index.ts` resolves per-launch HTTP config; per-provider `*/mcpComputerUse.ts` builders wire claude/codex/gemini/opencode + all ACP providers; threaded through the refactored `threadSession/` spawn pipeline (start/resume/recovery, structured + terminal + ACP).
- **Renderer**: composer opt-in (chip + add-menu + `@computer` mention), disabled for WSL; localized across all 12 catalogs.

## Problems found & decisions
1. **Refactor gap** — PR's launch plumbing lived in the old monolithic `threadSessionManager`; re-expressed in the new `threadSession/*` modules mirroring `browserMcp`, full parity across start/resume/recovery + ACP + CLI-hook.
2. **Security (P0)** — the computer-use token leaked into every child process env (opt-in was cosmetic). Scoped it: stripped from the base spawn env, injected per-launch only when `config.computerUse === true`; all 5 providers verified. Also: platform-gate the ingress (no Linux exposure), constant-time token compare, Host-header allow-list.
3. **Activation reliability** — bare `SetForegroundWindow` is refused by the Windows foreground-lock, silently dropping all input. Replaced with `AllowSetForegroundWindow(ASFW_ANY)` + alt-key + `AttachThreadInput` + retry, and **verify foreground before sending input** (fail fast). Probe-validated.
4. **`SW_RESTORE` un-maximized windows** → wrong click coords. Only restore if minimized; re-capture rect after activation.
5. **Input correctness** — `press_key` typed lowercase (dropped VkKeyScan shift bit); emoji surrogate-pair fragmentation; non-DPI-aware; stuck modifiers on crash; coordinate coercion to (0,0). All fixed and live-tested.
6. **macOS parity** (flagged for on-device verification) — `click` ignored `mouse_button` (silent left-click), `scroll` dropped `scrollX`, Retina pixel/point mismatch, geometry-hashed window id changed on move.
7. **Agent ergonomics** — removed two always-failing tools (`set_value`/`perform_secondary_action`) and never-populated accessibility fields; documented coordinate origin + window-refresh recovery; `get_window` now requires `app`.
8. **Speed** — the driver recompiled the inline C# shim on every call (~300ms of ~425ms/action). Replaced with a **long-lived PowerShell host** (compile once, NDJSON over stdin, correlated request queue, per-request timeout + kill/respawn, crash recovery, disposal on teardown). **Steady-state ~75ms/action (~40× faster on cheap ops).** Screenshots gained `max_dimension` (1280) + `format` (jpeg default): a large capture went **9.5MB → 80KB (99.2% smaller)**.

## Live-test evidence (real Win11 host)
Activation reaches a different-process window; `press_key "A"`→`A`, `!`→`!`, `ctrl+a` select-all; `ab😀cd` emoji intact; `launch_app` resolves the new window; UNC/URL launch rejected; persistent host serves concurrent NDJSON requests and recovers from a mid-session kill.

## Deferred (documented)
Per-capture consent/allow-list for sensitive windows (product UX decision); full UIAutomation/AX tree (current tree is an honest placeholder); browser/chrome token env scoping (same pre-existing pattern, out of scope); macOS on-device verification.

## Known limitation
Activating a few specific apps (Character Map, store Notepad, Windows Terminal) destroys their window handle at the OS level — confirmed **pre-existing** `Activate-Window` behavior, not introduced here. Ordinary apps are unaffected.

## Verification
`tsc` 0, `lint` 0, i18n 12/12 catalogs 0-missing, **4661 tests passing**. Windows driver behaviors live-tested with screenshots.

### DIFF
--- a/src/main/browser/BrowserMcpIngress.ts
+++ b/src/main/browser/BrowserMcpIngress.ts
@@ -1,60 +1,44 @@
-import { createServer, type IncomingMessage, type Server, type ServerResponse } from "node:http";
-import { randomBytes, randomUUID } from "node:crypto";
-import { decodeThreadIdentity, type McpThreadIdentity } from "@/shared/browserMcpThread";
-import { isLocalhostOrigin, readBoundedNodeRequestBody, writeJsonResponse } from "@/shared/http";
+import type { McpThreadIdentity } from "@/shared/browserMcpThread";
 import type { BrowserPanelManager } from "./BrowserPanelManager";
+import {
+  StreamableHttpMcpIngress,
+  type StreamableHttpMcpIngressInfo,
+} from "../mcp/StreamableHttpMcpIngress";
 import {
   BROWSER_MCP_INSTRUCTIONS,
   TOOLS,
   dispatchTool,
   formatToolResult,
   isKnownToolName,
-  type McpToolResult,
   type ToolContext,
 } from "./mcp/toolRegistry";
 
-export interface BrowserMcpIngressInfo {
-  url: string;
-  token: string;
-  port: number;
-}
-
-const MAX_BODY = 1024 * 1024;
-const MCP_PROTOCOL_VERSION = "2025-03-26";
-
-interface JsonRpcRequest {
-  jsonrpc: "2.0";
-  id?: number | string | null;
-  method: string;
-  params?: unknown;
-}
-
-interface JsonRpcResponseOk {
-  jsonrpc: "2.0";
-  id: number | string | null;
-  result: unknown;
-}
-
-interface JsonRpcResponseErr {
-  jsonrpc: "2.0";
-  id: number | string | null;
-  error: { code: number; message: string; data?: unknown };
-}
-
-type JsonRpcResponse = JsonRpcResponseOk | JsonRpcResponseErr;
+export type BrowserMcpIngressInfo = StreamableHttpMcpIngressInfo;
 
 /**
  * Single in-process MCP server. Speaks Streamable-HTTP MCP at `POST /mcp`
  * (JSON-RPC body, single JSON response). All five agent providers connect
- * here by URL — no per-thread Node child process.
+ * here by URL — no per-thread Node child process. The transport is shared
+ * with the computer-use ingress via {@link StreamableHttpMcpIngress}.
  */
 export class BrowserMcpIngress {
-  private server: Server | null = null;
-  private token = randomBytes(32).toString("hex");
-  private info: BrowserMcpIngressInfo | null = null;
   private allowEval = false;
   private allowDataAccess = false;
   private getManager: (() => BrowserPanelManager | null) | null = null;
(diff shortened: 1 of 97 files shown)


SDSLeon:
### Final integration/regression review — clean

An independent whole-diff review (distinct from the per-dimension correctness/speed/security/ergonomics/cross-platform passes) traced the integration seams and found **no material regression**. Verified safe:

- **Persistent-host lifecycle**: `before-quit → ingress.dispose() → driver.dispose() → child.kill()`; on a hard main crash the child self-terminates when its stdin pipe breaks (`ReadLine()` → null), so no orphan. Respawn-after-dispose supported.
- **Token scoping**: stripping `LIGHTCODE_COMPUTER_USE_MCP_*` from `sanitizedProcessEnv` is safe — `readComputerUseMcpEnv()` reads `process.env` directly, so opted-in resolution still works; per-launch injection is symmetric across all 5 providers (incl. the Codex `bearer_token_env_var` + `buildCodexComputerUseMcpEnv` pair).
- **Shared ingress**: `isAllowedHost` accepts localhost + IP literals (what every real caller sends; WSL via gateway IP), so browser behavior is unchanged; `timingSafeEqual` is functionally equivalent to the prior compare.
- **Framing/concurrency**: multi-MB base64 splits correctly (no `\n` in base64); timeout closures can only reject their own request id.

### Non-blocking follow-ups (documented, deliberately not fixed here)
1. **Linux composer** still shows the Computer Use toggle; resolves to a silent no-op (ingress not started on Linux) — strictly better than before the platform-gate, but the toggle could be hidden on unsupported platforms.
2. **Shared-host 20s timeout** recycles the singleton host and rejects any concurrent thread's in-flight request. Self-heals on next request; serialization is semantically required (SendInput is global). Acceptable tradeoff.
3. **Dead interface methods** — `set_value`/`perform_secondary_action` are removed from the tool surface but the `ComputerUseDriver` interface + driver impls still declare them (they throw if ever called directly). Harmless; can be pruned later.

PR #648, opened 2026-08-29
fix: renderer crash on empty pipelines, Cursor tool-schema rejection, dead SDK update

playtix-brain:
Three independent fixes, each with a regression test. All reproduced on 1.6.5/1.6.6 before fixing.

### 1. Renderer crash: `TypeError: e is not iterable`
`parseHeadFileView` indexed `parts[parts.length - 1]` without guarding an empty pipeline. A blank or pipe-only command (`bash -c ''`, `cd x && `, `|`) makes `splitShellPipeline` return `[]`, so `splitShellWords(undefined)` iterates `undefined` and takes down the whole renderer — the React error boundary replaces the app shell. It is reached from ordinary tool-call rows, so one odd command from an agent kills the UI.

### 2. Cursor turns fail with `NonRetriableError: Provider Error`
`spawn_agent` and `wait_for_agent` declared `oneOf` at the **root** of their `inputSchema`. Cursor's backend rejects root-level unions and fails the entire turn, for every model (reproduced with Fable 5, Sonnet 5 and Kimi K3). Verified outside the app with `cursor-agent acp` plus a minimal HTTP MCP server mirroring these tools: with `oneOf` the turn errors, without it the same prompt answers normally. The prompt/tasks and run_id/run_ids choices stay documented in the tool descriptions and are already enforced by the request parser.

### 3. Cursor SDK update action was dead
`cursorSdkUpdateCommand`/`canUpdateCursorSdk` only accepted `global-npm`/`global-pnpm`, which `sdkPackageDiscovery` reports only from the deferred `npm root -g` probe — and that probe never runs once a filesystem candidate matched. A Node prefix of its own (~/.local, nvm, fnm, volta, Homebrew) resolves as `global-inferred`, so no command was produced and the click fell through to the agent updater, refreshing the CLI while `@cursor/sdk` stayed on its old version. The update banner then never cleared.

Tests: 177 passing across the touched modules; `typecheck` clean on the changed files.


src/supervisor/crossagentMcp/toolRegistry.ts
```diff
@@ -226,12 +226,9 @@ const RAW_TOOLS: ToolSpec[] = [
           description: TIMEOUT_S_DESCRIPTION,
         },
       },
-      // Keep every root union branch explicitly object-typed. Some OpenAI-compatible
-      // providers reject `required`-only branches because they also match non-objects.
-      oneOf: [
-        { type: "object", required: ["prompt"] },
-        { type: "object", required: ["tasks"] },
-      ],
+      // No root-level union here: Cursor's backend rejects tool schemas that carry
```
SDSLeon, line 229: The root-level `oneOf` removal fixes Cursor compatibility, but the new comment says the request parser still enforces the choice and it currently does not.

- With both `prompt` and `tasks`, `spawnAgent` takes the `tasks` branch and silently discards `prompt`.
- With both `run_id` and `run_ids`, dispatch waits only for `run_ids` and silently ignores `run_id`.

Because the published schema now permits both shapes, a model-generated call can spawn or wait on the wrong work. Please keep the schema union-free, add explicit exactly-one validation before branch selection for both pairs, and add regression tests that pass both conflicting fields, assert an `isError` result, and verify that no spawn/wait manager method was called.

SDSLeon (changes requested):
Requesting changes for one blocking protocol-correctness issue. Removing the root-level unions is appropriate for Cursor compatibility, but the either/or contract must still be enforced at runtime so malformed model calls cannot silently select the wrong operation.

src/supervisor/crossagentMcp/toolRegistry.ts
playtix-brain: Done in 630e0c21 (merged with the branch updates). Dispatch now rejects calls carrying both `prompt` and `tasks`, and both `run_id` and `run_ids`, with an `isError` result before any run-manager call — and the schema comment no longer overstates what the parser did. Both pairs are covered by regression tests asserting the error and that no spawn/wait manager method is invoked.

SDSLeon (approved):
Follow-up verified: both-branch calls are rejected before any run-manager call, schemas stay union-free, regression tests cover both pairs. Tests, typecheck, and lint pass on the head. Thanks for the contribution!

PR #649, opened 2026-08-29
feat(antigravity): add GUI presentation mode

playtix-brain:
Antigravity is currently terminal-only (`presentationModes: ["terminal"]`), so it cannot be used as a regular chat like the other providers.

This adds a print-mode session adapter over the `agy` CLI's `stream-json` protocol and enables `gui` alongside `terminal`. Scope is limited to `src/supervisor/agents/antigravity/` — no shared runtime or UI code changes.

Verified end to end against `agy` 1.1.22: a GUI thread starts, streams and completes normally. Module tests: 62 passing.

### DIFF
--- a/src/supervisor/agents/antigravity/antigravity.test.ts
+++ b/src/supervisor/agents/antigravity/antigravity.test.ts
@@ -220,6 +220,19 @@ describe("createAntigravityAdapter", () => {
     expect(adapter.buildAcpLogoutCommand).toBeUndefined();
   });
 
+  it("creates a GUI structured session via agy print mode", async () => {
+    const adapter = createAntigravityAdapter();
+    expect(adapter.capabilities.presentationModes).toContain("gui");
+    const session = await adapter.createStructuredSession?.({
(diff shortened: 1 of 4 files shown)


SDSLeon:
Thanks for the contribution and for validating that Antigravity can work in Poracode's Chat presentation.

Since this was opened, Google has published an official Antigravity ACP server—the same `antigravity-acp` distribution Zed installs from the ACP Registry. Poracode already consumes that registry and has shared ACP session infrastructure, so we are going to implement built-in Antigravity Chat through Google's official ACP route instead of maintaining a provider-specific `agy -p` one-shot session.

That route gives us the native streaming/session lifecycle, resume, cancellation, tool updates, authentication, and a supported install/update path. Once the built-in integration is complete, we will also hide the separate Antigravity ACP registry entry to avoid presenting it as a duplicate provider.

I'm closing this PR in favor of that implementation. Thanks again for taking the time to build and test this—it helped establish the Chat use case and prompted us to investigate the now-available official integration.

PR #660, opened 2026-08-31
fix(claude): render MessageDisplay hook output

### LINKED ISSUE #659: bug: render Claude MessageDisplay-transformed responses
## OS

Windows 11

## Poracode version

v1.6.6-nightly.202608300831

## Agents involved

Claude Code (structured/GUI runtime) with a `MessageDisplay` hook

## Reproduction steps

1. Configure a Claude Code `MessageDisplay` hook that replaces or appends to assistant text.
2. Start a Claude GUI thread in Poracode.
3. Send a prompt and wait for the response to finish.
4. Observe that Poracode continues to show the original streamed text instead of the hook-transformed text from the final `assistant` snapshot.

## Expected behavior

Poracode should show the live stream while Claude is responding, then display the authoritative hook-transformed snapshot when the assistant item completes, without creating a duplicate message.

## Actual behavior

The Claude canonical mapper intentionally skips snapshot text when the same message was already streamed (`skipTextSnapshot`) to prevent duplication. Because `MessageDisplay` transformations are delivered only in the final `assistant` snapshot, this also discards the transformed text. The renderer then continues to prefer the original stream after completion.

## Additional context

Snapshot block indexes are not stable when blocks such as extended thinking are omitted, so the completed snapshot needs to be correlated with the existing streamed item using Claude's stable message ID rather than the content-block index.

SpookySandwich:
# Summary

Claude's `MessageDisplay` hook transforms an assistant response for display only. Poracode first receives the original text through streaming events, then the transformed text in the final `assistant` snapshot (`src/supervisor/agents/claude/canonicalMapping/dispatch.ts`).

The mapper now marks that snapshot payload `displayAuthoritative` when — and only when — it differs from the text that actually streamed. The field is optional on `messageItemPayloadSchema`; persisted items from before this change and stream-first providers carry no flag, so they keep the pre-existing stream-first behaviour (no version bump needed — recorded inline at the schema).

Every reader of assistant text now resolves it through one shared helper, `assistantDisplayText` / `hasVisibleAssistantText` (`src/shared/assistantMessageText.ts`): the chat row, the timeline visibility filter, find-in-chat, the handoff and experiment transcripts, MCP `read_thread`, and subagent run output. A flagged payload wins once the item completes — including intentionally empty text, which hides the row and keeps the suppressed original out of every export. In-flight and unflagged items keep showing the live stream, so Codex partial-payload turns and whitespace boundary chunks render exactly as on master.

# Motivation

Fixes #659. The first revision of this PR made the payload authoritative for every provider, which collapsed interrupted Codex answers to their first chunk, desynced find-in-chat from the rendered text, leaked hook-suppressed text into exports, and let a whitespace-only payload hide a real message. The marker confines the new behaviour to the payloads that actually need it.

# Testing

- [x] `pnpm run typecheck`
- [x] `pnpm run lint` (touched files: type-aware oxlint + oxfmt, also enforced by the pre-commit hook)
- [ ] `pnpm run fmt:check`
- [ ] `pnpm run test`
- [ ] Not run; reason:

Focused suites pass: 311 tests across the Claude canonical mapper, the shared helper, chat-pane selectors, `AssistantMessage`, find-in-chat, transcript exports, MCP thread tools, `SubagentRunManager`, and the runtime-event contract. New cases cover the review findings: interrupted Codex-style turns keep the stream, find matches the visible text, suppressed text stays out of exports, a whitespace payload can't hide real streamed text, and an untransformed snapshot emits no update. The full `pnpm run test` still shows the pre-existing Windows-only failures (ACP timing limits, absolute PowerShell/CLI path expectations) on this machine; the changed suites pass under Node 24.19.0 and pnpm 12.1.0.

# Screenshots

Not applicable: message lifecycle behaviour, no visual styling changes. The live-to-final transition and suppression cases are covered by renderer tests.

# Linked issue

Closes #659


src/renderer/components/thread/ChatPane/parts/items/AssistantMessage.tsx
```diff
@@ -53,14 +53,18 @@ export const AssistantMessage = memo(function AssistantMessage({
...
+      ? payloadTextBlocks
+          .map((block) => block.text)
           .filter(Boolean)
-          .join("\n") ?? "");
+          .join("\n")
+      : undefined;
   const isStreaming = item.state !== "completed";
+  // Keep live output responsive, then use the final payload as the display
+  // source of truth. An explicit empty text block intentionally hides the stream.
+  const rawText = isStreaming && stream.length > 0 ? stream : (payloadText ?? stream);
```
SDSLeon: **Blocking - this makes the payload authoritative for every provider, but only Claude emits an authoritative payload.**

Codex writes an assistant text payload from a different source than its stream. `codex/canonicalMapping/dispatch.ts:256-266` seeds a **partial** payload at `item/started` while `content.delta` keeps growing the stream. If the turn is then interrupted, `turn/completed` (`dispatch.ts:106-112`) and `forceCompleteTurn` (`codex/acp.ts:875-882`) emit `item.completed` with **no payload**, so that stale partial survives and now wins the render.

Measured in the running app, same injected sequence on both sides:

| | base | this PR |
|---|---|---|
| `CODEXPARTIAL first chunk.` rendered | yes | yes |
| `CODEXREST the full remainder...` rendered | yes | **no** |

The user watches the full answer stream in, then on completion the row silently collapses to the first chunk. Persisted Codex history is re-interpreted the same way on reload.

The same line also splits two readers this PR does not touch, because they still resolve assistant text stream-first while this renders payload-first:

- `chatFindMatches.ts:67` - for a hook-rewritten row, `getChatItemSearchText` returned `"RAWORIGINALWORD streamed body"` while the row rendered `"HOOKEDWORD replacement body"`. Searching the visible text gives **0 matches**; searching text that is nowhere on screen gives **1 match** that scrolls to the row and highlights nothing.
- `handoffTranscript.ts:44` and `experimentResponseTranscript.ts:57` use `payload || stream`. An intentionally-empty payload is falsy, so a hook that suppresses output hides the text on screen but **still exports the raw original** into an agent handoff. I confirmed both exports contain the suppressed marker while the UI does not.

**Proposed fix:** mark the payload as display-authoritative at the Claude boundary and gate this swap on that marker, so unmarked provider payloads keep losing to the stream. Then extract the precedence rule into one exported helper - `assistantDisplayText(item): string | undefined` - and use it here, in `isVisibleRuntimeItem`, in `getChatItemSearchText`, and in both transcript builders. The exports additionally need `payloadText !== undefined ? payloadText : stream` so that `""` is honoured rather than treated as falsy.

src/renderer/components/thread/ChatPane/chatPaneSelectors.ts
```diff
@@ -451,13 +451,20 @@ export function isVisibleRuntimeItem(item: RuntimeChatItem): boolean {
...
-    const hasPayloadContent = payload?.content.some(
-      (block) => (block.kind === "text" && hasVisibleText(block.text)) || block.kind === "image",
-    );
-    if (!(hasVisibleText(item.streams.assistant_text ?? "") || hasPayloadContent)) return false;
+    const hasPayloadText = payload?.content.some((block) => block.kind === "text") ?? false;
+    const hasPayloadContent =
+      payload?.content.some(
+        (block) => (block.kind === "text" && hasVisibleText(block.text)) || block.kind === "image",
+      ) ?? false;
+    const hasVisibleContent = hasPayloadText
```
SDSLeon: **Blocking - a whitespace-only payload now hides a message that has real streamed content.**

`hasPayloadText` is true for *any* `kind === "text"` block, so the payload becomes solely decisive and the stream is no longer consulted at all. A completed item with payload `[{kind:"text", text:"\n\n"}]` plus a non-empty stream was visible before and is dropped now.

Measured in the running app: base renders the message and `isVisibleRuntimeItem` returns `true`; on this PR it returns `false` and the row never reaches the DOM. The content is simply gone from the transcript.

This is reachable outside Claude - `codex/canonicalMapping/dispatch.ts:339` gates the final payload on `finalText.length > 0`, not on *visible* text, which is the same `"\n\n"` stream-boundary shape the comment just above cites for Factory Droid. The existing `payload-whitespace` fixture only covers the no-stream case, so it does not catch this.

**Proposed fix:** gate on the same display-authoritative marker as `AssistantMessage`, so only a payload the provider actually vouches for can override a non-empty stream. While here, the ternary flattens with identical behaviour and drops the duplicated operand:

```ts
const hasVisibleContent =
  hasPayloadContent || (!hasPayloadText && hasVisibleText(item.streams.assistant_text ?? ""));
```

Worth adding a fixture with a whitespace payload **and** a non-empty stream to lock this in.

src/supervisor/agents/claude/canonicalMapping/dispatch.ts
```diff
@@ -415,14 +416,36 @@ function mapClaudeSdkMessageInner(
...
         if (!block || typeof block !== "object") continue;
         const obj = block as Record<string, unknown>;
-        if (obj.type === "text" && typeof obj.text === "string" && obj.text.length > 0) {
-          if (skipTextSnapshot) continue;
+        if (obj.type === "text" && typeof obj.text === "string") {
+          if (skipTextSnapshot) {
+            const streamedItem = streamedTextItems[streamedTextItemIndex];
+            streamedTextItemIndex += 1;
+            if (streamedItem) {
+              events.push({
```
SDSLeon: **Non-blocking - this fires on every message, including when nothing was transformed.**

The update is emitted unconditionally whenever `skipTextSnapshot` is true, so an ordinary turn with no `MessageDisplay` hook configured still emits an `item.updated` whose text is byte-identical to what already streamed. `main/db/runtimeItems.ts:600-612` persists the merged payload and preserves `row.streams`, so from here on every Claude assistant turn's full text is stored twice in SQLite.

Confirmed with a mapper probe: a snapshot identical to the streamed text still produces the `item.updated`.

**Proposed fix:** retain the accumulated streamed text on `TextItemState` and skip the update when the snapshot text matches it. That is also a useful narrowing for the renderer concerns above - with no hook installed, nothing downstream would ever see a divergent payload.

SDSLeon (changes requested):
## Request changes

The fix works - I verified that. But the renderer half of it changes behaviour for **every** provider, and that introduces three regressions plus one export leak. All four are reproducible in the running app against a base-commit control.

### How I verified

Two isolated Electron sessions from the same fixture - one at the merge base (`21c834c9f`), one at this PR head (`3948dbe91`) - driven through the real store, reducer, selectors, components, and the real `chatFindMatches` / transcript modules. Identical injected event sequences in both. The sequences were reconstructed from each provider's mapper source, not captured from a live provider process; that is the one gap in this verification.

| Observation | base | this PR | |
|---|---|---|---|
| Hook-rewritten text rendered | absent | **shown** | :white_check_mark: #659 fixed |
| Codex: full streamed answer rendered | shown | **missing** | :x: regression |
| find-in-chat matches the *visible* text | 0 | **0** | :x: regression |
| find-in-chat matches text *not on screen* | 1 | **1** | :x: phantom match |
| Whitespace payload + real stream rendered | shown | **hidden** | :x: regression |
| Suppressed text present in handoff/experiment export | true | **true** | :x: now a leak |

### Two concerns I withdrew

Flagging these so nobody chases them:

1. **Snapshot/stream block-count mismatch.** I initially thought positional correlation could drop or duplicate text. It cannot. The CLI's completed-message rewrite maps over `content`, so block count is preserved, the full transformed text goes into the *first* text block and later text blocks are blanked to `""` - exactly the shape your `["First display block", ""]` test asserts. Across 316 real transcripts (49,586 assistant records) there are zero empty text blocks and zero genuinely-multi-text messages. The design is right.
2. **The `isReplay` change.** Also right, and better justified than the PR description says: in those same transcripts, 272 message ids re-emit their entire block sequence byte-for-byte. Preserving the frame is what dedupes them. I could not reproduce any mid-block replay problem.

### Why this is "request changes" rather than a nit

The Codex regression needs no hook, no Claude, and no unusual input - just an interrupted turn - and it silently truncates an answer the user already watched stream in. It also re-interprets already-persisted history on reload. That is the blocker; the rest travel with it.

### Suggested direction

The root cause is that the change introduces a *second* source of truth for the same text without marking which one wins. Two shapes:

- **Marker (smaller, recommended):** flag the authoritative payload at the Claude provider boundary and gate the renderer swap on that flag, then route all five readers through one shared `assistantDisplayText(item)` helper. This adds an optional field to a persisted payload, so it is a compatibility boundary under `.agents/docs/versioning.md`. It happens to be safely backward compatible - older items have no flag, so they fall back to the stream, i.e. base behaviour - but worth making that an intentional call.
- **Replace the stream instead (root fix):** if the mapper could overwrite `assistant_text` rather than write a competing payload, none of these four findings could exist and the renderer would need no change at all. `src/shared/contracts/runtimeEvent.ts` only has the appending `content.delta` today, so this means a new event type plus reducer and persistence changes - a real version audit. Bigger, but it removes the whole class.

Your call on which - you may know constraints I do not.

Minor, not blocking: `textItems.ts:20` ("New messages clear the map on `message_start`, so a fresh frame will see no `existing`") is now false for same-id replays, which is the path this PR relies on. And the `it.each` tuple carries both `_order` and `stopBeforeSnapshot`, which encode the same axis and force the conditional assertions.

SpookySandwich:
Thanks for the thorough review — the base-commit control harness made all four findings easy to confirm. Everything is addressed in the new commit (`2900e642`), point by point:

**1. Payload-first arbitration regressed other providers (Codex interrupted turns, find-in-chat, export leak).** Agreed — the root cause was exactly what you named: a second source of truth with nothing marking which one wins. I took the marker shape you recommended. The Claude mapper now sets an optional `displayAuthoritative` flag on the payload, and only when the final snapshot actually differs from what streamed. The renderer swap is gated on that flag, so unmarked payloads keep losing to the stream — Codex interrupted turns and pre-flag persisted history render exactly as on master. All five readers you listed go through one shared helper (`assistantDisplayText` / `hasVisibleAssistantText` in `src/shared/assistantMessageText.ts`), and the exports honor `""` through it, so suppressed text no longer leaks. I also routed two more readers of that same text through the helper — MCP `read_thread` and subagent run output — since they had the same leak; both are no-ops for unmarked items.

**2. Whitespace-only payload hiding real streamed content.** Fixed by the same gate: the visibility filter now consults the helper, so only a payload the provider vouches for can override a non-empty stream. Added the fixture you asked for — whitespace payload plus a non-empty stream, unflagged, stays visible (`unflagged-whitespace-payload-with-stream`) — plus the flagged counterpart that correctly hides.

**3. Unconditional `item.updated` on untransformed snapshots.** Done as you suggested: `TextItemState` retains the accumulated streamed text, and the mapper skips the update when the snapshot matches it. With no hook installed, no event fires and nothing is double-persisted — and as you noted, nothing downstream ever sees a divergent payload in that configuration.

**4. Versioning call.** Intentional, per `.agents/docs/versioning.md`: the field is optional, old data stays valid, and every reader falls back to the stream when it's absent — i.e. base behaviour. That reasoning is recorded inline at the schema rather than as a version bump or migration.

**Minor items.** The stale `textItems.ts` comment now describes the same-id replay path it actually serves, and the `it.each` tuple was split into two tests (before/after block completion) so the redundant axis encoding is gone.

**Your two withdrawn concerns.** Noted and agreed — no code chases either path. A snapshot block without a streamed lane is treated as unreachable per your transcript data, and a late snapshot after a frame reset still dedupes to nothing (locked in by a test).

Verification on my side: 311 focused tests pass (mapper, helper, selectors, `AssistantMessage`, find, both exports, MCP thread tools, `SubagentRunManager`, contract schema), plus typecheck and type-aware lint. The full suite still shows the same pre-existing Windows-only failures (ACP timing limits, absolute PowerShell/CLI path expectations) — untouched by this PR.

SDSLeon (approved):
Verified the revised provider-gated MessageDisplay behavior and compatibility path. Approving this head; remaining reader edge cases will be fixed separately on master.

PR #686, opened 2026-09-03
feat(handoff): scale transcript budget to the destination context window

playtix-brain:
Stored-history handoffs used a fixed 50,000-character budget, discarding useful earlier conversation even when the destination had a much larger context window. Desktop context files now scale to approximately 35% of the destination window (four characters per token), while preserving the original ask and prioritizing recent conversation over tool activity.

- 200k context: 280,000 characters; 1M context: 1,400,000 characters. Unknown sizes use 400,000 characters. File context is capped at 4,000,000 characters for remote attachment delivery.
- Small windows receive proportionally smaller transcripts, with message truncation that retains the original request and includes header/gap costs in the budget.
- Numeric context parsing is shared with the existing settings parser, including case, commas, decimals, and 1k–10M bounds.
- Mobile uses the destination budget capped at 50,000 characters. Desktop attachment failures use the same inline ceiling, retaining the beginning and end with an omission marker. This accommodates the remote 1 MiB JSON request limit, including duplicated prompt/segments and escaped text.

This affects stored-history `context-file` handoffs. Provider-side extraction and `thread-transcript` handoffs continue through their existing paths. No persisted or wire shape changes are required.

Validation: 115 targeted tests passed across transcript selection, desktop dialog wiring, mobile switch/fork delivery, attachment success/failure, shared parser compatibility, and handoff strategy. Typecheck and touched-file standard/type-aware lint and formatting checks passed.


SDSLeon (commented):
Deep review (findings-only, PR tests verified 11/11 locally). Direction looks right — scaling the ceiling to the destination window instead of a fixed 50k is the correct fix for front-truncation eating the goal/constraints. Details inline.

Scope question to confirm: since gui→gui handoffs now take the `thread-transcript` route (own thread on switch, thread mention on fork — nothing copied), this PR only affects `context-file` handoffs: anything involving a terminal, mirrored threads, `read_thread` disabled, and the mobile path. Is that the intended readership? Related: 35% of the destination window rides the new session's first message for the rest of its run — deliberate trade vs. a smaller share?

src/renderer/actions/handoffTranscript.ts
```diff
@@ -7,7 +7,46 @@ import { formatHandoffRow, type HandoffRow } from "./handoffTranscriptRows";
  * window, but the file rides in the new provider's first message for the rest
  * of its session, so it is filled by priority rather than recency alone.
  */
-export const MAX_TRANSCRIPT_CONTEXT_CHARS = 50_000;
+export /**
```
SDSLeon: The `/** Whole-file budget, roughly 12-15k tokens… */` block above (lines 5–9) documented the deleted `MAX_TRANSCRIPT_CONTEXT_CHARS` and now contradicts the new comment below it (“Small next to any current context window”). It was orphaned when the constant was removed — suggest deleting lines 5–9.

SDSLeon: Nit: the leading `export` on this line now exports `HANDOFF_CONTEXT_SHARE` — leftover from replacing `export const MAX_TRANSCRIPT_CONTEXT_CHARS`. Only `handoffTranscriptBudget` needs to be public; suggest dropping this `export`.

src/renderer/actions/handoffTranscript.ts
```diff
@@ -7,7 +7,46 @@ import { formatHandoffRow, type HandoffRow } from "./handoffTranscriptRows";
...
+ * The budget now scales with the destination model's context window and only
+ * claims a share of it, leaving the rest for the actual work.
+ */
+const HANDOFF_CONTEXT_SHARE = 0.35;
+const CHARS_PER_TOKEN = 4;
+/** Used when the destination declares no context size. */
+const DEFAULT_MAX_TRANSCRIPT_CONTEXT_CHARS = 400_000;
+
+/** `"1m"` / `"200k"` / `"272000"` -> token count, or undefined when unparsable. */
+function tokensFromContextSize(contextSize: string | undefined): number | undefined {
```
SDSLeon: Nit: this reimplements `parseContextWindowInput()` (`src/shared/agents/codexContextWindows.ts`) for the same `contextSize` id format, minus its comma handling and 1k–10M clamps — e.g. `"20m"` is rejected there but yields a ~28M-char budget here. Suggest reusing the shared parser and keeping the 400k floor on top.

src/renderer/actions/handoffTranscript.ts
```diff
@@ -80,6 +119,7 @@ function joinRows(rows: readonly HandoffRow[], kept: ReadonlySet<HandoffRow>): s
 export function buildTranscriptContext(
   thread: Thread,
   sourceLabel: string,
+  maxChars: number = DEFAULT_MAX_TRANSCRIPT_CONTEXT_CHARS,
```
SDSLeon, line 99: Coupling note: raising this default 50k → 400k silently 8×'s the other caller, `src/mobile/useRemoteDesktop.ts:1184`, which passes no budget — and mobile inlines the transcript into the prompt with no attachment-file fallback (the `buildHandoffLaunchInput` inline path), so it is the path most exposed to prompt-size limits. `input.targetConfig.contextSize` is in scope there; suggest threading `handoffTranscriptBudget(...)` through it too, or calling out mobile as intentionally default-only.

src/renderer/actions/handoffTranscript.test.ts
```diff
@@ -205,10 +209,30 @@ describe("buildTranscriptContext", () => {
...
+    const summary = buildTranscriptContext(thread, "Claude", TEST_BUDGET)?.summary ?? "";
 
     expect(summary).toContain("[turns omitted]");
     // The header line rides outside the row budget, hence the small slack.
-    expect(summary.length).toBeLessThanOrEqual(MAX_TRANSCRIPT_CONTEXT_CHARS + 500);
+    expect(summary.length).toBeLessThanOrEqual(TEST_BUDGET + 500);
+  });
+});
+
+describe("handoffTranscriptBudget", () => {
```
SDSLeon, line 230: Suggestion: nothing tests the PR's core claim — that the dialog forwards the *destination's* `contextSize` — and there are no cases for realistic ids (`"1M"`, `"272k"`, `"default"`, `"1.05M"`). A wiring test plus parametrized ids would lock this (and the parser-parity point on `tokensFromContextSize`) in.

SDSLeon (changes requested):

SDSLeon:
@playtix-brain please fix and respond to comments, thank you!

SDSLeon (approved):
Reviewed the adopted changes at 638ca74d6 and resolved the five prior review threads. No remaining Important findings after behavior, integration, compatibility, and simplification review.

Destination-based sizing is covered through the desktop dialog; mobile and failed-file delivery stay within the existing remote JSON limit. Small contexts retain the original ask, parser behavior remains compatible with saved settings, and successful attachment delivery preserves the larger transcript.

Validation: 115 targeted tests passed, including long Unicode/escaped payloads and mobile switch/fork paths. Typecheck plus touched-file standard/type-aware lint and formatting passed. GitHub CI is running and will be checked before merge.

PR #738, opened 2026-09-07
Add Copy image and Save image to enlarged previews

SpookySandwich:
# Summary

Right-click an enlarged image to copy it to the system clipboard or save the original file. Works for input attachments and generated images, preserves zoom and gallery selection, and lets Escape dismiss the menu before closing the preview.

The preview and thumbnail actions share image loading and clipboard conversion. Original filenames and formats stay attached to pasted images and remote gallery URLs. Desktop HTTP image reads use the bounded main-process transport, and JPEG copies convert to PNG for browser clipboard compatibility.

# Motivation

Images could already be copied or downloaded from image cards, but those actions were unavailable while viewing an enlarged image.

# Testing

- [x] `pnpm run typecheck`
- [x] Both lint modes on all changed TypeScript files
- [x] Formatting check on all changed TypeScript files
- [x] 100 focused tests across image actions, preview menus, image sources, galleries, and HTTP transport
- [x] Localization extraction: zero missing strings in all 12 non-English catalogs
- [x] Isolated Electron mock smoke: baseline, thread search, and all four selected mock gates pass; zero console/runtime errors; test app stopped
- [x] Native Windows checks: PNG and WebP copied as images; saved PNG SHA-256 matches its original

The full repository test suite and external provider sessions were not run for this focused UI change.

The optional binary response encoding only changes local renderer/main IPC; existing text requests keep their default behavior, covered by the existing HTTP tests. No persisted state or remote peer protocol changes.

# Screenshots

Enlarged image Copy and Save menu

# Linked issue

None.

### DIFF
--- a/src/main/ipc/localHandlers.remoteHttpRequest.test.ts
+++ b/src/main/ipc/localHandlers.remoteHttpRequest.test.ts
@@ -106,6 +106,19 @@ describe("local remoteHttpRequest handler", () => {
     ).resolves.toMatchObject({ status: 204 });
   });
 
+  it("preserves binary image responses when base64 is requested", async () => {
+    const bytes = new Uint8Array([0x89, 0x50, 0x4e, 0x47, 0, 0xff, 0x80]);
+    vi.stubGlobal(
+      "fetch",
+      vi.fn<FetchMock>(async () => new Response(bytes)),
+    );
+    const result = await makeHandlers().remoteHttpRequest({
+      url: "https://remote.example.test/api/files/image?path=sample.png",
+      responseEncoding: "base64",
+    });
+    expect(new Uint8Array(Buffer.from(result.body, "base64"))).toEqual(bytes);
+  });
+
   it("rejects non-http protocols before fetching", async () => {
     const fetchMock = vi.fn<typeof fetch>();
     vi.stubGlobal("fetch", fetchMock);
--- a/src/main/ipc/localHandlers.ts
+++ b/src/main/ipc/localHandlers.ts
@@ -298,7 +298,7 @@ export function createLocalIpcHandlers(
         return {
           status: response.status,
           headers: headersToRecord(response.headers),
-          body: Buffer.from(buffer).toString("utf8"),
+          body: Buffer.from(buffer).toString(payload.responseEncoding ?? "utf8"),
         };
       } catch (error) {
         if (controller.signal.aborted) {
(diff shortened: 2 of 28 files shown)


SDSLeon:
@SpookySandwich Hi, Good idea! Thank you! 

What do you think about instead of context menu put them at the bottom near zoom controls, like icons? Maybe both. Just thinking that Context menu here is not so clear for users.

SDSLeon (approved):
Reviewed and verified locally on `b0b7a916` (checked out, `pnpm run typecheck`, oxlint on all changed files, focused vitest on the six touched suites — 59/59 green, `pnpm i18n:extract` shows 0 missing in every locale and is a no-op on the committed catalogs).

What I checked beyond the tests:

- **IPC boundary**: `responseEncoding` is a zod-validated, opt-in optional on `remoteHttpRequest`, which is a `MAIN_LOCAL_PROCEDURE` — it never crosses the remote peer protocol, and the existing `mainProcessFetch` caller keeps the default `utf8`. Response body stays bounded by `readBoundedResponseBody`, so base64 can't balloon memory. No versioning concern.
- **exactOptionalPropertyTypes**: conditional spreads where values are genuinely optional; `ImageViewSource.mime/fileName/alt` are required strings, so the direct passthrough into `LightboxImage` is sound.
- **Behavior changes are deliberate and tested**: JPEG now re-encodes to PNG for clipboard compatibility (previously passed through raw), undecodable images fail loudly instead of silently copying raw bytes, and single-image cards now route through the gallery so original filenames survive opaque remote URLs.
- Clean dedup: `imageActions.ts` / `imageUrlMetadata.ts` pull shared logic out of `ImageCard`/`imageViewSource` rather than duplicating it in the lightbox, and the lightbox menu reuses the shared `ContextMenuSurface` (viewport-aware positioning, backdrop dismissal) instead of inventing a new popover. Escape/arrows are correctly handed to the menu while it's open.

Nice work — the metadata preservation through opaque remote URLs was the subtle part and it's well covered by tests.

SpookySandwich:
> @SpookySandwich Hi, Good idea! Thank you!
> 
> What do you think about instead of context menu put them at the bottom near zoom controls, like icons? Maybe both. Just thinking that Context menu here is not so clear for users.

sure! I will add the icons. I was copying the behavior from claude, but having icons is definitely clearer.

SpookySandwich:
Thanks for merging! I pushed the follow-up to the same branch: [2e04175f](https://github.com/SpookySandwich/lightcode/commit/2e04175f457feb54a19a33a1feb46094ed91f93d).

Copy and Save are now buttons beside the zoom controls, reusing the image-card icons with localized tooltips. They replace the custom context menu. Since this PR was already merged, the follow-up commit needs to be applied separately.

Rechecked locally: 73 tests across six focused suites, typecheck, both lint modes, formatting, and localization (0 missing; extraction unchanged) all pass. Electron smoke also passed with zero console/runtime errors.

Copy and Save buttons beside zoom controls

PR #744, opened 2026-09-07
feat: add a follow-up queue alongside Steer

### LINKED ISSUE #737: feat: support Queue vs. Steer for follow-up messages
### Problem

I want to send messages while the agent is running, but currently there is no clear distinction between Queueing (run after current turn finishes) and Steering (interrupt or redirect right away). A typical key-binding can be adapted, e.g. enter for queue and ctrl+enter for steer. Keys can be reversed in settings.

### Proposed solution

When queueing, Poracode holds the message in a local queue and automatically sends it once the active turn finishes. When steering, Poracode stops the active turn and immediately dispatches the new message.

### Alternatives considered

A potentially cleaner approach is to check the support for each provider and use their native interface whenever possible.

### Additional context

_No response_

SpookySandwich:
Queued messages wait for the current chat turn and any open requests to finish, then run in FIFO order. Settings > Threads chooses whether Enter steers or queues; existing users keep Steer, while Ctrl/Cmd+Enter selects the opposite action.

The queue appears above the composer with a left drag grip and Steer, Edit, and Remove controls on hover or keyboard focus. Touch devices keep those actions visible. Reordering uses stable item IDs so concurrent queue changes cannot resurrect removed messages. Editing and Stop pause delivery; Resume continues it. Edits preserve attachments and mentions, retain local text after concurrent changes, and reject stale saves.

The supervisor owns queue state and mirrors it to desktop and mobile clients. Queues survive renderer and remote reconnects and remain in memory for the supervisor session.

Direct Steer waits for pending queued or replacement turn admission before choosing a submission path. Its reservation stays held through input preparation and asynchronous native provider acceptance. If the old turn ends during preparation, the fallback start establishes a fresh completion boundary before FIFO can advance. A resumed session whose first submission rejects now settles the current replacement with an error state and a visible runtime error, while stale failures leave newer sessions untouched.

Closes #737.

## Integration with #747

The combined code from this PR and #747 at `d761240e` has now been tested with the real SDK adapter and supervisor, replaying SDK messages. **The two PRs are not compatible as written.** After the SDK accepts a native follow-up, the preceding turn's result makes the adapter report idle. FIFO then calls `startTurn` for its next item before the native reply has finished, taking turn ownership prematurely.

`sdkSession.followUpQueue.test.ts` reproduces this: it expects one `startTurn` call after the preceding result and observes two. The test is dormant while native steering is absent and activates automatically when #747's capability is integrated. A provider-neutral regression also verifies that FIFO correctly waits across this gap when the provider preserves working status.

The `StructuredSessionHandle.steerTurn` contract now documents the requirement: resolving the promise acknowledges acceptance; an adapter with accepted work queued internally must retain working status across the preceding completion, open the next canonical turn for the reply, and publish idle only after that work finishes. #747 must honor this before the two PRs can be merged together. Its existing slash-command mapper and permission-mode review findings also remain outstanding at the tested head.

Additional local checks: 72 tests passed across the relevant SDK, queue, and admission suites; the native-capability integration test is intentionally skipped on #744 alone and fails on the combined checkout. The SDK suite passed 33/33 in isolation after an error-path timeout under concurrent load. Typecheck, both lint passes, formatting, and commit hooks passed. This is deterministic integration coverage, not a live authenticated provider run.

## Validation

- Two Luna Max revision passes completed for admission/steering and restart failure handling.
- Local runtime validation: 224 passed, 1 skipped across 19 suites. Thread view and composer validation: 94 passed across 2 suites, including both follow-up buttons and submission actions while Stop is pending.
- After separating the admission regressions into their own suite, all 19 admission, queue-manager, and restart tests passed across 3 suites.
- Six new regressions fail against the original PR commit `c796c818` and pass with the revision.
- Typecheck, regular and type-aware touched-file oxlint, formatting, whitespace checks, and commit hooks passed.
- [CI for the updated head](https://github.com/Porabuild/Poracode/actions/runs/34401953793) requires repository-admin approval before it can run.

The initial implementation's validation included localization extraction with 0 missing translations in all 12 non-English locales, deterministic Electron smoke, and live Codex checks for FIFO, Steer, Edit/Resume, Remove, Stop/Resume, reorder, command-only completion, and reconnect recovery. This runtime revision adds no user-facing strings. Physical mobile pairing and live approval/question dialogs were not exercised. 

## Compatibility

Remote v9 remains additive: new procedures and optional queue snapshot state, with explicit unsupported errors on older hosts. Existing settings normalize to Steer. Edit concurrency tokens are optional for older clients. The revision changes only in-memory runtime coordination and error handling; persisted shapes and wire contracts remain valid, so no additional version bump or migration is needed.

## Screenshot

Queued follow-ups with left reorder grips and Steer, Edit, and Remove controls

Queue actions appear on hover or keyboard focus; the left grip supports reordering.


src/supervisor/runtime/threadSessionManager.ts
```diff
@@ -960,25 +1152,67 @@ export class ThreadSessionManager {
...
+  ): Promise<void> {
+    const directInput = this.followUpQueue.beginDirectInput(payload.threadId);
+    let targetSession: SessionRuntime | undefined;
+    let admittedSession: SessionRuntime | undefined;
+    try {
+      // A provider switch/reconnect can leave the old SessionRuntime in the
+      // map while its replacement owns the serialized start lock. Wait before
+      // preparing or admitting input so a late rejection cannot reset the new
+      // session's direct barrier.
+      await this.waitForPendingStart(payload.threadId);
```
SDSLeon, line 1201: [P1] Wait for queued/replacement admission before admitting Steer

Unlike sendThreadInput, this entry point does not wait for waitForQueuedTurnAdmission or waitForPendingSteerAdmission. If queued A has called startTurn but is still asynchronously preparing while status is idle, clicking Steer on queued B reaches markStarted, whose onStarted hook removes A's active ownership, then starts B as another normal turn. This is reachable with ACP: runStartTurn awaits sessionConfigSync.applyTurnConfig before canonical start/working and before promptInFlight protects the call. Two prompts can therefore overlap admission and corrupt turn ownership.

Apply both admission barriers before deciding how to submit the steer. Add a regression that delays A's admission while idle, steers B, and asserts no second startTurn until A is admitted. Important; high confidence.

src/supervisor/runtime/threadSessionManager.ts
```diff
@@ -960,25 +1152,67 @@ export class ThreadSessionManager {
...
+      }
+
+      const effectiveConfig = applyHomeScopePermissions(
+        effectiveProjectLocation(session),
+        payload.config,
+        session.adapter.capabilities,
+      );
+      session.config = effectiveConfig;
+      admittedSession = session;
+      directInput.markStarted(session);
```
SDSLeon, line 1238: [P1] Keep the direct-input reservation through steer preparation

markStarted decrements the reservation here, but SteerCoordinator.setPendingSteer still awaits attachment rewriting and resolveSkillTurnInjection before submitting anything. With a native-steer provider, directAwaitingReplacement is false. If the current turn completes and goes idle during that preparation, maybeFinishDirect clears the direct barrier and the FIFO pump can submit its next item first; the user's eventual steer is then delivered to that queued turn instead of retaining priority. Queue-row promotion has an extra outer reservation, but ordinary composer steering does not.

Retain the reservation until the actual prepared-input handoff (or move the mark into a post-preparation admission hook). Test with deferred skill preparation, completion of the old turn, and an existing FIFO item; that item must not start before the steer is submitted. Important; high confidence, with the early barrier release reproduced against the actual direct-input implementation.

src/supervisor/runtime/threadSession/spawnPipeline.ts
```diff
@@ -1002,14 +1006,7 @@ export class SpawnPipeline {
...
         };
-        void structuredSession
-          .startTurn(prompt, launchConfig, turn.segments, startOptions)
-          .catch((error) => {
-            if (ctx.sessions.get(restarted.threadId)?.instanceId !== restarted.instanceId) {
-              return;
-            }
-            ctx.failStructuredSession(restarted, error);
-          });
+        return await structuredSession.startTurn(prompt, launchConfig, turn.segments, startOptions);
```
SDSLeon: [P1] Report restart submission failures on the newly spawned instance

spawnThread above replaces the SessionRuntime, but this now propagates startTurn rejection to restartThreadSettlingFailure, whose catch only reports errors when the old session is still current. That condition is false after a successful spawn. For example, resuming a Claude GUI thread with a deleted/unreadable image makes startTurn emit working and then reject from buildSdkUserMessage's readFile, without a provider onError event. The new session remains working without the runtime error item; the previous catch explicitly failed the restarted instance.

Retain the new instance and, if it is still current, failStructuredSession on it before rethrowing for queue recovery. Add a regression where spawning succeeds but the replacement's first startTurn rejects without emitting onError, asserting both errored status and a runtime error item. Important; high confidence.

SDSLeon (changes requested):
Reviewed the full PR at c796c8183 with separate lifecycle, composer/reuse, and transport/compatibility passes. Requesting changes for three confirmed runtime issues detailed inline: Steer bypasses pending admission, direct-steer preparation releases FIFO protection too early, and resumed start failures no longer settle the replacement session with an error.

Validation: traced the real ACP asynchronous configuration path and Claude attachment-read failure path; a focused Node probe using the PR's actual FollowUpQueueDirectInput implementation confirmed that its barrier can clear before steer preparation finishes. No additional confirmed UI or transport findings. Local Vitest could not start because dependency bootstrap failed on missing electron/package.json; direct invocation also found the Vitest module unavailable. GitHub format/lint/typecheck passed, but test shard 5 failed in ThreadView.test.tsx (missing “Send message” label); I have not attributed that CI failure to one of the inline findings.

No source fixes or deferred refactors in this findings-only review. I also reviewed #747; after resolving both PRs, cover a native SDK follow-up spanning an earlier turn's completion with another FIFO message pending.

SDSLeon (dismissed):
## Deep review — Approved
Revalidated PR #744 at 9a29e284: no validated Important findings. Safe to merge from review; not merging per policy.

### What's good
- threadSessionManager.ts:1167 now joins queued/replacement admission before Steer chooses its delivery path.
- followUpQueueDirectInput.ts:38 retains reservations through preparation and acceptance, while fresh fallback starts reset completion ownership. The awaited native handoff in steerCoordinator.ts closes the asynchronous admission gap.
- spawnPipeline.ts:1009 settles a failed restart on the current replacement runtime, preserving error state and the runtime error item without mutating a newer replacement.

### Coverage
Revalidated all three original findings against the revision delta, lifecycle callers and regression coverage, with independent read-only review and primary validation. Runtime suites passed 112 tests with two skips; ThreadView.test.tsx additionally passed all 29 tests, including the previously failing surface. Full typecheck, ordinary/type-aware lint and formatting on all 11 revision files, and diff whitespace checks passed. Required ignored Codex protocol bindings were generated before typecheck. No live authenticated provider run was performed; GitHub reports no CI checks.

The native Claude FIFO integration case is capability-skipped here because native steering belongs to unmerged #747. This approval applies to #744 independently; #747 remains changes-requested and its combined FIFO behavior must pass before that PR is approved.

SDSLeon:
@SpookySandwich please resolve conflicts

PR #747, opened 2026-09-09
feat(claude): steer an in-flight turn instead of interrupting it

playtix-brain:
Sending a message while Claude is working currently interrupts its running turn. This change accepts follow-ups without interrupting foreground Bash or subagents, then delivers them at the next natural turn boundary.

The Claude adapter keeps accepted steers locally until the preceding result. This preserves live tool mapping, applies model/permission/fast/ultracode configuration through the normal turn-start path, and keeps the thread working so the ordinary FIFO queue cannot overtake a steer. Slash commands use the same boundary. Stop, force completion, disposal, and turn failures discard pending steers; generation checks suppress cancelled asynchronous delivery and stale errors. User rows retain a stable ID across admission and delivery.

Preserves the original contributor's commits and adds maintainer fixes for the review findings and follow-up queue integration. All production changes stay inside the Claude provider; existing event and persisted data shapes remain compatible, so no version bump is needed.

Validation:
- Claude and supervisor suites: 783 passed, 2 skipped.
- Full application typecheck and touched-file lint/type-aware lint/format checks passed.
- Live authenticated Claude CLI test passed: foreground Bash finished and both replies completed, with no intermediate idle state.
- Regression coverage includes slash commands, tool output/errors, multiple steers, permission transitions, live settings, FIFO ordering, Stop during asynchronous preparation, stale errors, and failed attachments.

Live test: `PORACODE_LIVE_CLAUDE_STEER=1 pnpm exec vitest run --configLoader runner --config vitest.integration.config.ts tests/integration/claude-steer.integration.test.ts`


src/supervisor/agents/claude/sdkSession.ts
```diff
@@ -320,6 +321,40 @@ export class ClaudeSdkSession implements StructuredSessionHandle {
...
+   */
+  async steerTurn(
+    prompt: string,
+    config: ThreadConfig,
+    segments?: PromptSegment[],
+    options?: StartTurnOptions,
+  ): Promise<void> {
+    if (this.disposed) return;
+    if (!this.currentTurnInFlight || prompt.trimStart().startsWith("/")) {
+      return this.startTurn(prompt, config, segments, options);
```
SDSLeon: [P1] Do not reset a live turn's mapper for slash-command steering

Sending /compact while foreground Bash is running immediately calls startTurn, even though the old SDK turn has not stopped. startClaudeTurn clears the streaming maps and removes ordinary foreground tools from toolItemsById. When that Bash command returns, canonicalMapping/dispatch.ts looks up tool_use_id, finds no entry, and skips its tool_result, losing command output and its success/error result. The live-subagent exception does not preserve ordinary Bash tools. The new slash test only checks for turn.started and misses the lost result.

Serialize the command until the current turn ends, or use the coordinated interrupt path before resetting turn state. Add a Bash tool_use → steer /compact → tool_result regression that verifies the output and failure/success update survive. Important; high confidence.

src/supervisor/agents/claude/sdkSession.ts
```diff
@@ -320,6 +321,40 @@ export class ClaudeSdkSession implements StructuredSessionHandle {
...
+    }
+    this.currentConfig = config;
+    // No `turn.started`: the in-flight turn keeps its lifecycle. Only the user
+    // row is painted, with a stable id so a later fallback to `startTurn`
+    // re-emits the same (deduped) item.
+    this.emitRuntimeEvents(
+      steerClaudeTurn(this.mapperState, prompt, segments, options?.userMessageItemId),
+    );
+    const query = await this.requireQuery();
+    await this.syncModel(query, config);
```
SDSLeon: [P1] Apply the submitted permission mode before queuing native steer

This path only syncs the model; it skips the permissionModeForConfig/setPermissionMode block used by startTurn. If the user switches a running thread from agent/bypassPermissions to Plan and submits a follow-up, currentConfig and the UI now say Plan, but the SDK continues using its previous permission mode when it consumes the queued message. Conversely, leaving Plan through a mid-turn submission remains stuck in the SDK's old mode. Assigning currentConfig alone does not update that SDK control.

Share the applicable live-config synchronization with startTurn and apply the requested permission mode before pushing the message (also preserve the existing fast/ultracode synchronization). Test mid-turn transitions into and out of Plan and a changed approval policy by asserting the SDK control call precedes enqueue. Important; high confidence; the exact new method was probed with a plan-mode config and made no permission-mode call.

SDSLeon (changes requested):
Reviewed all four changed files at d761240e, including the SDK lifecycle, canonical tool-result mapping, configuration synchronization, and tests. Requesting changes for the two confirmed issues inline: slash commands reset the running turn's mapper, and native steering omits the permission-mode synchronization performed by startTurn.

Validation: traced foreground Bash tool registration through tool_result handling and executed the exact new steerTurn method in a focused Node probe, confirming that a plan-mode submission enqueues without calling setPermissionMode. Existing beginResumedTurnIfNeeded does reopen the lifecycle on later assistant activity, so I did not treat the absence of an immediate turn.started for ordinary steering as a separate defect. No concrete simplification findings or deferred refactors.

Full Vitest/runtime verification remains unperformed because local dependencies are incomplete; GitHub reports no checks for this branch. No source edits. Please also validate with #744: a pending native SDK follow-up must remain ahead of FIFO delivery across the preceding result/idle boundary.

SDSLeon (dismissed):
Reviewed the adopted implementation at ac37c0267, including the targeted lifecycle re-review. The two original review findings are fixed: slash commands wait until tool results are mapped, and follow-ups use normal model/permission/session-flag synchronization. The provider-local queue also prevents premature FIFO delivery and makes pending steers cancellable by Stop; generation guards cover both late submission and stale failures.

Validation: 783 Claude/supervisor tests passed (2 existing skips), full app typecheck and touched-file lint/format passed, and a live authenticated Claude test preserved foreground Bash output and completed both replies without an intermediate idle state. Stable user IDs preserve transcript deduplication. Production changes remain provider-local with no incompatible persisted or wire shape changes. No validated Important findings remain.

PR #748, opened 2026-09-09
fix(codex): render citations from preinstalled OpenAI artifact skills

SpookySandwich:
# Summary

Codex replies produced by OpenAI's preinstalled artifact skills can display raw `:codex-file-citation{...}` directives in Poracode. Render these as clickable file chips in saved and new replies, preserving complete Windows paths and correctly identifying PDF, PowerPoint, Word, and spreadsheet references as files.

The Codex provider parses its directive syntax through the existing transcript-formatting hook. CommonMark structure determines code boundaries, including nested list and blockquote containers; an iterative walk handles deep nesting. Full Markdown and its lazy fallback share a versioned file-link codec that encodes paths separately from line metadata, so filenames ending in `:2026` or `:20-26` open the correct file. Legacy links remain readable.

Generic autolinking preserves existing URL identity while the project index loads, and path normalization retains external UNC host prefixes. Literal directives in code and incomplete or malformed directives remain intact at the provider boundary.

# Motivation

OpenAI ships the **PDF skill** with the ChatGPT/Codex primary runtime. The runtime manifest includes the artifact plugin bundle, and the PDF plugin is preinstalled and enabled in the affected setup. Routine PDF creation can therefore produce this markup without a user separately installing a skill or choosing an add-on.

The PDF skill (`26.905.11957`, "Final response citations") explicitly requires the model to cite generated PDFs using `:codex-file-citation{path="..." purpose="output"}`. The affected session read that instruction before emitting the directive. The bundled document and spreadsheet skills require the same format; the presentation skill recommends it when the host app supports it.

Any Poracode user running Codex with these bundled skills can encounter the issue during ordinary artifact workflows. Supporting their output is part of compatibility with OpenAI's supplied runtime. This impact is scoped to Poracode's rendering of that runtime's output; it does not establish identical skill distribution across every standalone CLI or ChatGPT surface.

Poracode did not interpret that syntax. It displayed the directive as text, and ordinary path detection could turn only the trailing portion of a Windows path into a folder chip. The observed session used GPT-6 Astra, but the evidence identifies the skill instruction as the source of the format; it does not establish an Astra-specific regression.

Existing transcript data remains valid and requires no migration. The change uses the existing file viewer; Office previews and page, slide, or cell navigation are outside its scope. Provenance and documentation references are recorded in `.agents/docs/codex-file-citations.md`.

# Testing

- [x] `pnpm run typecheck`.
- [x] `pnpm run lint`.
- [ ] `pnpm run fmt:check` — flags the unchanged `CLAUDE.md` symlink, materialized as target text in this Windows checkout. All changed files pass formatting.
- [ ] Previous full-suite run at `c27ac02d`: 11,023 passed, 12 failed, 64 skipped. Rerunning all seven failing files with two workers produced 224 passes and two failures in unchanged supervisor code: the Windows resume-command assertion in `runtime.test.ts:1637` and Volume GUID skill filtering in `SkillsService.test.ts:530`. This focused revision reruns the affected suites below.
- [x] 137 targeted tests across 10 suites, including all reviewer repros, full/fallback opening callbacks, legacy link compatibility, multiline attributes, 5,000 nested blockquotes, and UNC/WSL path normalization. The original reviewer regressions first reproduced as nine failing renderer cases on the previous implementation.
- [x] Isolated Electron smoke: baseline, thread search, and all four selected mock gates passed; zero console/runtime errors during the smoke run.
- [x] Manual Electron checks: three artifact chips and full-path tooltips; reviewer cases for list fences, unmatched backticks, blockquote indented code, and colon-suffix filenames; saved-history reload with raw text unchanged; a citation opening a fixture file through the real editor/IPC path. The earlier verification also covered streaming completion.
- [x] Two Luna Max implementation workflows and an independent Luna Max review. The additional UNC and deep-nesting findings are addressed and regression-tested.

External provider turns, network-share access, and Office previews were not exercised. UNC callback preservation is tested without accessing a live share. The isolated Electron harness needed a temporary longer startup allowance under concurrent host load; the harness was restored and every session was stopped. File opening was verified by activating the actual DOM button because CDP mouse input did not dispatch a click on this host.

# Screenshots

Real Electron app, same viewport and reply, with sample paths. Before uses `ce33db77`; after shows the revised rendering. Screenshots are hosted on a separate evidence branch in the contributor fork.

| Before | After |
| --- | --- |
| Raw citation directives in the original renderer | Artifact citations rendered as file chips |

Reviewer regressions: code stays literal, later citations render, and colon suffixes remain part of the filename.

Review regression examples in the Electron app

# Linked issue

Reported directly with a screenshot; no issue number.

### DIFF
--- a/src/renderer/components/providers/codex/fileCitationMarkdown.ts
+++ b/src/renderer/components/providers/codex/fileCitationMarkdown.ts
@@ -1,0 +1,104 @@
+import { getBasename } from "@/shared/pathUtils";
+import { pathRefUrl } from "../../thread/ChatPane/parts/items/markdownPathRefs";
+
+const CITATION_START = ":codex-file-citation{";
+
+/**
+ * Codex artifact skills emit file directives in assistant text, including saved
+ * transcripts. Normalize them at the provider rendering boundary, before generic
+ * path autolinking can split Windows paths or guess an artifact is a folder.
+ * See .agents/docs/codex-file-citations.md for provenance and supported syntax.
+ * No persisted text is changed. Incomplete/malformed directives stay literal.
+ */
+export function formatFileCitationMarkdown(text: string): string {
+  if (!text.includes(CITATION_START)) return text;
+  const tokens = /^ {0,3}(?:> ?)*(`{3,}|~{3,})[^\r\n]*|`+|:codex-file-citation\{/gm;
+  let fence: string | undefined;
+  let inlineTicks = 0;
+  let cursor = 0;
+  let result = "";
+  for (let match = tokens.exec(text); match; match = tokens.exec(text)) {
+    const token = match[0];
+    if (match[1]) {
+      const marker = match[1];
+      if (fence) {
+        if (
+          marker[0] === fence[0] &&
+          marker.length >= fence.length &&
+          token.slice(token.indexOf(marker) + marker.length).trim() === ""
+        ) {
+          fence = undefined;
+        }
+      } else if (!inlineTicks) {
+        fence = marker;
+      }
+      continue;
+    }
+    if (fence) continue;
+    if (token[0] === "`") {
+      if (inlineTicks === token.length) inlineTicks = 0;
+      else if (!inlineTicks && !isEscaped(text, match.index)) inlineTicks = token.length;
+      continue;
+    }
+    if (inlineTicks || isEscaped(text, match.index)) continue;
+    const lineStart = text.lastIndexOf("\n", match.index - 1) + 1;
+    if (/^(?: {4}|\t)/.test(text.slice(lineStart, match.index))) continue;
+    const citation = readCitation(text, tokens.lastIndex);
+    if (!citation) continue;
+    result += text.slice(cursor, match.index) + citation.markdown;
+    cursor = citation.end;
+    tokens.lastIndex = cursor;
+  }
+  return cursor ? result + text.slice(cursor) : text;
+}
+
+function isEscaped(text: string, index: number): boolean {
+  let slashes = 0;
+  while (text[--index] === "\\") slashes++;
+  return slashes % 2 === 1;
+}
+
+function readCitation(text: string, start: number): { markdown: string; end: number } | null {
+  // Quoted attribute values may contain spaces, braces, or Markdown punctuation.
+  // This is a directive, not JSON: preserve native Windows backslashes.
+  const attribute =
+    /\s*([a-zA-Z_][\w-]*)\s*=\s*(?:"((?:\\.|[^"\\\r\n])*)"|'((?:\\.|[^'\\\r\n])*)')/y;
+  const fields = new Map<string, string>();
+  let cursor = start;
+  while (cursor < text.length) {
+    const end = /^\s*\}/.exec(text.slice(cursor));
+    if (end) {
+      const path = fields.get("path");
+      // The artifact skills specify absolute filesystem paths. Never turn an
+      // arbitrary URL/scheme into a local-file action.
+      if (
+        !path ||
+        !/^(?:[a-zA-Z]:[\\/]|\/|\\\\)/.test(path) ||
+        // eslint-disable-next-line no-control-regex -- Control characters are invalid in file references.
+        /[\u0000-\u001f\u007f]/.test(path)
+      ) {
+        return null;
+      }
+      const label = getBasename(path).replace(/[\\`*_[\]<>|!]/g, "\\$&");
+      try {
+        return {
+          markdown: `[${label}](${pathRefUrl({ kind: "file", path })})`,
+          end: cursor + end[0].length,
+        };
+      } catch {
+        // Malformed Unicode in provider text must not take down the chat row.
+        return null;
+      }
+    }
+    attribute.lastIndex = cursor;
+    const match = attribute.exec(text);
+    if (!match || fields.has(match[1]!)) return null;
+    const quote = match[2] === undefined ? "'" : '"';
+    const value = (match[2] ?? match[3]!).replace(/\\(["'])/g, (raw, escaped: string) =>
+      escaped === quote ? escaped : raw,
+    );
+    fields.set(match[1]!, value);
+    cursor = attribute.lastIndex;
+  }
+  return null;
+}
--- a/src/renderer/components/providers/codex/fileCitationMarkdown.ts
+++ b/src/renderer/components/providers/codex/fileCitationMarkdown.ts
@@ -1,7 +1,42 @@
+import { unified } from "unified";
+import remarkParse from "remark-parse";
 import { getBasename } from "@/shared/pathUtils";
 import { pathRefUrl } from "../../thread/ChatPane/parts/items/markdownPathRefs";
 
 const CITATION_START = ":codex-file-citation{";
+const CITATION_START_RE = /:codex-file-citation\{/g;
+
+interface MarkdownNode {
+  type: string;
+  children?: MarkdownNode[];
+  position?: MarkdownPosition;
+}
+
+interface MarkdownPosition {
+  start?: { offset?: number };
+  end?: { offset?: number };
+}
+
+interface SourceRange {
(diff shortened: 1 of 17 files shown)


src/renderer/components/providers/codex/fileCitationMarkdown.ts
```diff
@@ -0,0 +1,104 @@
...
+          marker[0] === fence[0] &&
+          marker.length >= fence.length &&
+          token.slice(token.indexOf(marker) + marker.length).trim() === ""
+        ) {
+          fence = undefined;
+        }
+      } else if (!inlineTicks) {
+        fence = marker;
+      }
+      continue;
```
SDSLeon, line 148: [P2] Keep Markdown code state from leaking past a list-contained fence

A valid response such as the following leaves the final real citation completely unconverted:

````text
- ```text
  example
  ```

Created :codex-file-citation{path="/tmp/report.pdf" purpose="output"}
````

The list opener misses the fence regex and sets `inlineTicks = 3`; the indented closer matches the fence branch, but `!inlineTicks` is false and this `continue` prevents clearing the inline state. Every subsequent directive is skipped at line 43. I reproduced this with the actual formatter source. Please recognize Markdown containers/code boundaries so the example stays literal and the subsequent citation renders. Cover unmatched prose backticks too (they currently suppress all later citations), and blockquote indented code (`>     <directive>`, currently rewritten despite being literal code). Important, high confidence.

File: src/renderer/components/thread/ChatPane/parts/items/markdownPathRefs.ts [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    path = decodeURIComponent(encoded);
+  } catch {
+    path = encoded;
+  }
+  if (!path) return null;
+  if (prefix === AUTO_PATH_FOLDER_HREF_PREFIX || prefix === AUTO_PATH_FOLDER_PREFIX) {
+    return { kind: "folder", path };
+  }
+  const match = path.match(/^(.+):(\d+)(?:-(\d+))?$/);
```
SDSLeon, line 136: [P2] Preserve literal colon-number suffixes in explicit file paths

The new citation entry point accepts arbitrary absolute POSIX filenames, but `pathRefUrl({ kind: "file", path: "/tmp/report:2026" })` followed by this decoder returns `{ kind: "file", path: "/tmp/report", line: 2026 }`. Both renderers then open a different file (or fail to open it). Percent encoding cannot protect the suffix because this heuristic runs after decoding. The legacy sentinel ambiguity is newly exposed to explicit citation paths by this PR.

Encode explicit path and optional line metadata independently, retaining a compatibility branch for legacy links. Add roundtrip and click-callback regressions for filenames ending in `:2026` and `:20-26` with no line metadata. Important, high confidence; reproduced with the actual codec source.

SDSLeon (changes requested):
Request changes: the provider hook and shared link extraction are appropriate, but two reproducible correctness issues remain before this reliably renders artifact citations.

**Scope reviewed**
Reviewed all 13 changed paths at c27ac02d5062e3a437e565d47b0e7e0140a3516b against ce33db772f0e3a95bcae48874d919c923523e82a, including the parser, manifest registration, full Markdown and lazy fallback, path URL codec, tests, and compatibility/provenance documentation. Independent read-only passes covered parser behavior and shared rendering/reuse; the primary pass validated findings, contracts, and test evidence. Applied repository architecture, provider isolation, rendering, localization, and versioning rules. No code changes or fix re-review.

**Confirmed findings**
1. Important / high confidence: Markdown code-state tracking leaks after a list-contained fence, so subsequent genuine citations remain raw. Related context cases also mishandle unmatched prose backticks and blockquote indented-code examples. See the parser line comment for reproduction and the required regression cases.
2. Important / high confidence: explicit POSIX paths ending in colon plus digits are interpreted as line metadata, truncating the file passed to the opening callback. Keep explicit path identity separate from line metadata while retaining legacy links.

**Rejected after validation**
No provider-isolation violation: Codex syntax stays in its provider folder and uses the existing supplied hook. No persisted-state migration is required: raw transcript data is unchanged and old sentinel decoding remains. No localization additions are required for dynamic file basenames. No additional actionable simplification finding: extracting the codec and tokenizer removes existing duplication.

**Deferred major or dangerous refactors**
None.

**Verification**
- Read the full frozen diff and relevant callers, including both rendering paths and file-opening normalization.
- Checked installed OpenAI artifact skill instructions: the documented directive format is present in the PDF, document, and spreadsheet skills.
- Ran a Node diagnostic using the actual PR source with TypeScript types stripped: ordinary citation conversion succeeds; a list-contained fence and unmatched prose backtick leave later citations raw; blockquote indented code is rewritten; /tmp/report:2026 decodes to /tmp/report with line 2026.
- Attempted targeted Vitest suites for citation parsing/rendering, shared path tokens/autolinking, formatter registration, and existing Markdown rendering. Tests did not start: dependency bootstrap failed because electron/package.json was unavailable. A direct Vitest launch also found the local Vitest module unavailable. These are review-environment limitations, not attributed PR failures.
- Confirmed the PR head is unchanged before submission and the isolated checkout has no tracked edits.

**Blocked or unverified**
Automated suites and a fresh Electron interaction were not independently completed. The PR author's reported test/smoke results were reviewed as evidence, not represented as checks run by this review. The two requested changes are supported by direct source diagnostics and caller tracing.

src/renderer/components/providers/codex/fileCitationMarkdown.ts
SpookySandwich: Fixed in c0e561031. Markdown code/container boundaries now come from the existing CommonMark parser instead of persistent fence/backtick state. Complete directives are masked at stable source offsets so quoted attribute punctuation does not affect classification, and only text ranges are rewritten. The AST walk is iterative to handle deep nesting safely.

Regressions cover the exact list-contained fence, unmatched prose backticks, and blockquote indented-code examples, plus nested containers, multiline LF/CRLF attributes, and 5,000 blockquote levels. Full/fallback renderer tests verify the subsequent citations open correctly. All 137 targeted cases pass; the PR includes a fresh Electron screenshot of the reviewed examples.

src/renderer/components/thread/ChatPane/parts/items/markdownPathRefs.ts
SpookySandwich: Fixed in c0e561031. New internal links use v2/<encoded-path> with independent line/endLine query fields. Both legacy HTTPS and poracode sentinel forms retain their previous decoding behavior, including the malformed-percent fallback.

Roundtrip and full/fallback click regressions preserve /tmp/report:2026 and /tmp/report:20-26 without line metadata, while actual line ranges and stored legacy links still open correctly. Integration also caught and fixed heuristic reclassification of existing URLs before the project index loads, and independent Luna Max review caught UNC host-prefix loss, now fixed in shared path normalization. All 137 targeted cases pass, with typecheck, repository lint, and PR-file formatting green.

SDSLeon (approved):
## Deep review — Approved
Revalidated PR #748 at c0e561031: no validated Important findings. Safe to merge from review; not merging per policy.

### What's good
- CommonMark text-range classification in fileCitationMarkdown.ts fixes the reported list fence, unmatched backtick, and blockquote code cases without rewriting literal examples.
- markdownPathRefs.ts keeps explicit filename identity separate from line metadata while retaining legacy decoding. Full and fallback click coverage preserves colon-number filenames, and shared normalization retains UNC roots.

### Coverage
Reviewed the revision delta against the previous review, surrounding formatter/rendering/codec callers, provider isolation, compatibility and regression coverage, with independent read-only review and primary validation. 102 tests passed across eight suites. Full typecheck, ordinary and type-aware lint on changed TS/TSX files, formatting on all 17 PR files, and diff whitespace checks passed. Required ignored Codex protocol bindings were generated before typecheck. No fresh Electron interaction was run; GitHub reports no CI checks.

PR #750, opened 2026-09-10
feat(crossagents): persist routing fallback chains

### LINKED ISSUE #730: feat: Persistent fallback chains for Crossagents routing preferences
### Problem

Crossagents already has two powerful routing capabilities that don't currently compose:

1. Persistent, tag-based routing preferences (`set_routing_preference` / `crossagentRoutingOverrides`) — e.g. `#implementation + #backend → provider/model A`.
2. Per-call ordered fallback chains (`spawn_agent` `fallbacks` + `retry_on`) — e.g. `A → B → C` for that single invocation.

There is no way to persist the combination:

`#implementation + #backend → A → B → C`

This means users who want a "preferred workhorse first, escalate to stronger/scarcer models only if needed" strategy must re-type the fallback chain every time they delegate a task. It also prevents the Crossagents ranker from applying any learned or manual route to the fallback chain itself.

The goal is strictly additive: keep the existing ranking, manual overrides, and runtime fallback machinery exactly as they are, and let a routing preference optionally carry an ordered fallback chain.

### Proposed solution

Extend the existing routing-preference mechanism so that a persistent route can include an optional `fallbacks` array and an optional `retryMode`.

Conceptually:

```ts
{
  tags: ["implementation", "backend"],
  agentKind: "...",
  modelId: "...",
  fallbacks?: [
    { agentKind: "...", modelId: "..." },
    { agentKind: "...", modelId: "..." }
  ],
  retryMode?: "startup" | "any-failure"
}
```

At spawn time, if a routing preference matches the task tags and has fallbacks, the supervisor would expand it into a `SpawnAgentRequest` whose `fallbacks` are validated the same way per-call fallbacks are today (via `prepareSubagentRun` / `validateAgentModelSelection`).

I have not implemented anything yet; I'd like to align on design first.

Open questions:
- Is extending `crossagentRoutingOverrides` the right approach, or would a separate routing-policy type be preferable?
- Should resolution happen in `resolveSelectionArgs` (the roster-ranker layer) or closer to `prepareSubagentRun` / `spawnPlan.ts`?
- Would you prefer schema + resolver + tests in one PR, and UI / tool editing in a follow-up?

### Alternatives considered

1. **Keep supplying `fallbacks` manually on every `spawn_agent` call.**
   - Works today.
   - Does not persist user intent and cannot be triggered by learned/manual routing.

2. **Introduce a separate routing-policy subsystem.**
   - Could be more powerful, but risks duplicating the existing `crossagentRoutingOverrides` / learned-rank concepts already built in PRs like #408, #479, and #540.
   - I'd prefer to extend the current mechanism unless maintainers want a broader redesign.

3. **Cost/quota-aware automatic routing.**
   - Interesting future direction.
   - Explicitly out of scope for this request because provider cost/quota signals are not uniformly authoritative across agents.

### Additional context

I reviewed the current source in Poracode `1.8.0-nightly.202609061723` and the routing/fallback machinery appears to already provide most of what this feature would reuse: manual routing overrides, learned ranking, provider/model filtering, runtime fallback execution, and `retryMode` handling. The request is to make the composition of routing preferences and fallback chains persistent, not to replace the router.

Razikale365:
Adds persistent fallback chains to `crossagentRoutingOverrides`:

- `fallbacks` and `retryMode` are persisted with the routing preference.
- The saved chain is carried through `rankCrossagentCandidates` and applied during `spawn_agent` only when the primary matches the override.
- Per-call `fallbacks`/`retry_on` take precedence and `fallbacks: []` disables inheritance.
- Settings migrations (Qwen, Antigravity) and validation keep the feature backward-compatible.
- Batch spawns remain atomic: an invalid inherited fallback rejects the whole batch before any child starts.

Closes #730


src/supervisor/crossagentMcp/toolRegistry.ts
```diff
@@ -736,6 +790,47 @@ async function setRoutingPreference(
...
+        typeof fallback.provider === "string" && fallback.provider.length > 0
+          ? fallback.provider
+          : "";
+      if (!fallbackProvider) return errorResult(`fallbacks[${i}].provider is required`);
+      mapped.push({
+        agentKind: fallbackProvider,
+        ...(typeof fallback.model === "string" && fallback.model.length > 0
+          ? { modelId: fallback.model }
+          : {}),
+        ...(typeof fallback.reasoning === "string" && fallback.reasoning.length > 0
```
SDSLeon, line 819: **Important — data loss: validate the complete override before saving**

**What's wrong:** A valid primary with `fallbacks: [{ provider: "claude", reasoning: "x".repeat(257) }]` is accepted here and `set_routing_preference` returns `saved`. The persisted selection schema limits provider/model/effort to 256 characters. After this override is written, `readSharedSettingsFile` calls `normalizeSharedSettings`, which rejects the entire `crossagentRoutingOverrides` array and replaces it with `[]`; all existing routes disappear from loaded settings and subsequent writes can make that loss permanent. I reproduced the actual dispatch → upsert → serialization/normalization path: two routes before reload, zero afterwards.

**Why guards don't stop it:** `resolveSelectionArgs` validates only the primary; the fallback mapper checks nonempty strings but not schema bounds. Persistence and the settings writer do not validate this event payload.

**Fix:** Validate the fully constructed override with `crossagentRoutingOverrideSchema.safeParse` before calling `ctx.setRoutingOverride`, returning a tool error on invalid input without mutating persisted settings.

**Regression check:** Start with an existing valid route and submit an oversized fallback provider/model/reasoning field. Assert a tool error, no persistence call, and the original route retained after reload.

SDSLeon (changes requested):
## Deep review — Request changes
Scope: PR #750 at a2c7c85bb, all nine changed files. 1 Important — must fix before merge.

### Must fix
- src/supervisor/crossagentMcp/toolRegistry.ts:818 — fallback fields bypass the persisted schema, so one oversized value invalidates every saved route on settings reload. See inline.

### Verified good
- Matched manual routes carry their chains into the existing spawn planner. Explicit chains (including []) replace inheritance, and explicit retry policy wins.
- All complete batch plans are validated before any child starts; existing retry/cancellation behavior is retained.
- Legacy routes remain valid and nested retired model/alias migrations preserve fallback entries.

### Coverage and validation
Three independent read-only passes covered runtime/selection, persistence/ranking, and proof/public interface/systemic reuse; the primary review inspected the full executable diff and personally reproduced the finding. Applied repository architecture, provider-isolation and versioning instructions and issue #730 requirements. No source edits or targeted re-review.

- pnpm exec vitest run src/shared/settings.test.ts src/shared/settings.fallbacks.test.ts src/shared/crossagentRanking.test.ts src/supervisor/crossagentMcp: 231 tests passed across 10 suites.
- pnpm run typecheck: passed after required ignored protocol generation.
- Ordinary and type-aware oxlint, oxfmt on all nine changed files, and git diff --check: passed.
- GitHub CI: all reported checks passed, including eight test shards.
- Focused diagnostic bundled the actual PR source, called dispatchTool with a valid primary and a 257-character fallback reasoning value, then used the real upsert and settings normalizer: toolIsError=false, savedFallbackLength=257, beforeRead=2 routes, afterRead=0 routes.

### Rejected after validation
Top-level batch fallback/retry defaults are not a regression: the established batch interface limits inherited defaults to provider/model/reasoning/fast/permissions, while task-level fallback/retry settings are honored. No compatibility version bump is needed for optional fields that retain legacy parsing.

No live provider execution was performed. No deferred refactors.

src/supervisor/crossagentMcp/toolRegistry.ts
Razikale365: Fixed in 2c418b05. The fully constructed routing override is now validated against `crossagentRoutingOverrideSchema` before persistence. Invalid nested fallback fields return a tool error without calling the persistence layer. Added a regression covering the oversized fallback case and preservation of existing routes across reload.

SDSLeon (approved):
Re-reviewed the fix in 2c418b05 on head 1916f6f7: the single Important (persisted-schema bypass) is resolved — `set_routing_preference` now validates the fully constructed override via `crossagentRoutingOverrideSchema.safeParse` before persistence and returns a tool error without mutating settings, with a regression covering oversized nested fallbacks. Re-ran the routing/settings/Crossagents suites locally (233 passed). LGTM.

PR #759, opened 2026-09-12
fix(a11y): single-pane thread view is exposed as a disabled button

SpookySandwich:
## Problem

In the default single-pane layout the whole thread pane, composer included, is exposed to the operating system as a disabled button. Open DevTools → Accessibility on a fresh install: the pane container carries `role="button"`, `tabindex="0"`, `aria-roledescription="draggable"` and `aria-disabled="true"`. Chromium propagates the disabled state to every descendant, so screen readers, voice control and dictation tools report the composer and every control in the pane as unavailable.

## Cause

dnd-kit's accessibility plugin decorates a draggable's activator, `handle ?? element`. `ThreadPane` and `DraftPane` only attached the header drag handle when two or more panes existed, so a lone pane fell back to the pane element itself.

## Fix

- Always attach the handle.
- The handle now wraps only the pane title, so the element dnd-kit decorates never contains other controls or the content.
- The window drag-region class keys off the pane count instead of handle presence, so a lone pane's header still moves the window.

Multi-pane reordering is unchanged apart from the grab area being the title rather than the whole header row.

## Verification

- ThreadView, ThreadDraftView and ThreadComposerSection suites pass (131 tests). Typecheck, oxlint and oxfmt clean.
- Isolated dev build: the pane element carries no ARIA attributes, the only decorated element in the pane is the title handle, and no ancestor of the composer is `aria-disabled`.
- Windows UI Automation reports the composer as an enabled Edit control. Found via Typeless dictation, which stopped showing its fallback popup once the pane was no longer disabled.


SDSLeon (approved):
## Deep review — Approved

**Scope:** PR #759 @ `03a1736`, 5 files (`ThreadView.tsx`, `ThreadDraftChrome.tsx`, `ThreadDraftView.tsx`, `ThreadPane.tsx`, `DraftPane.tsx`) — no validated Important findings. Safe to merge from review; not merging per policy.

### What's good

- **The diagnosis matches the library.** `@dnd-kit/dom@0.5.0` resolves the activator as `draggable.handle ?? draggable.element` and then unconditionally adds `tabindex="0"`, `role="button"`, `aria-roledescription="draggable"`, `aria-describedby`, and syncs `aria-disabled` to `draggable.disabled` — for *every* registered draggable, disabled or not. With the handle attached only when `paneCount > 1`, a lone pane really did get the whole pane element decorated and `aria-disabled="true"`. Always attaching the handle is the correct minimal fix at this dnd-kit version.
- **The narrowed handle is the right shape.** `ThreadView.tsx:388-417` and `ThreadDraftChrome.tsx:41-51` now wrap only the title, so the injected `role="button"` no longer encloses `ProjectSwitchMenu`, continue-in-provider, mark-done, the close button, or the composer. Accessible name still resolves from content (`Tooltip.Trigger` is `role="none"`, which does not suppress descendant text).
- **Re-keying the drag region off `paneCount` is behaviour-preserving.** `ThreadPane.tsx` and `DraftPane.tsx` both previously gated `dragHandleRef` on exactly `paneCount > 1`, so `paneDraggable ? "poracode-content-over-drag-region" : "…--drag"` selects the identical class in every case. Swept the other call sites — `AppContent.tsx` (non-compact) and `QuickComposerOverlay.tsx` (`quickComposer`, `paneCount={1}`) never reach the compact header — so no surface flips its macOS window-drag behaviour.
- **Memoization is not disturbed.** `ThreadView`'s hand-written comparator checks `dragHandleRef` identity (`ThreadView.tsx:88`); always passing it could have defeated the memo, but `@dnd-kit/react`'s `handleRef` is a `useCallback` keyed on the stable draggable instance, so identity holds across renders. The comparator also covers `paneCount`, so the newly derived `paneDraggable` cannot go stale.
- **Layout is a pure pass-through.** The added wrapper in `ThreadView.tsx:385` inherits the old handle's exact classes and the handle takes the slot the `Tooltip.Trigger` held; flex free-space resolves identically, both `gap-2` runs survive, and `@container` stays on the unchanged row so the `@max-[560px]` / `@max-[360px]` title breakpoints still resolve against the same container.
- **Keyboard reordering survives the smaller activator.** `KeyboardSensor` binds `keydown` to `handle ?? element` and requires `event.target === handle`; the injected `tabindex="0"` makes the title div the focus target, the header is `app-region: no-drag` in multi-pane, and the inner trigger is `tabIndex={-1}` so it is not a competing tab stop.

### Coverage

Two review lanes over the full diff — (1) behaviour, lifecycle, and accessibility correctness, (2) layout, cross-surface parity, and repository conventions — plus direct spot-checks: dnd-kit's accessibility plugin and keyboard sensor source, the `.poracode-content-over-drag-region` / `--drag` rules in `styles.css`, every `ThreadView` / `ThreadDraftView` / `ThreadDraftCompactHeader` call site, the `ThreadView` memo comparator, and `paneCount` / `dragHandleRef` prop typing under `exactOptionalPropertyTypes`. Checked and cleared as pre-existing rather than introduced: the `experiment && !thread.worktreePath` placeholder branch in `ThreadPane.tsx` still leaves the pane element as the activator (it renders no header, and the decorated node holds only static text), and a lone pane's activator remains a focusable `aria-disabled="true"` node — previously the entire pane element, now just the title, so strictly improved. No user-facing strings added, so no Lingui obligation.

PR #764, opened 2026-09-13
fix(projects): prevent duplicate folders and preserve project data

SpookySandwich:
# Summary

Opening the same folder again now reuses its existing project. Different folders can still share a name, and projects on different machines stay separate.

Reopening a project keeps its custom setup script, actions, and settings. An explicit name or workspace selection is applied to the existing project. Desktop, remote, and MCP creation use the same identity rule, and moving a project to an already registered folder is rejected.

# Motivation

Each folder used to receive a new project ID whenever it was added. Preventing new duplicates also needs a safe upgrade for profiles that already have them.

The new database migration moves thread ownership before deleting duplicate rows, preserving conversation history. It also merges notes and to-do lists and repairs saved views, pane layouts, and experiments. Live project updates preserve unsent rich text and attachments, including two populated drafts in split panes. Home remains a separate synthetic project.

The repair keeps the oldest project, regardless of its sidebar position. Its non-empty settings win conflicts; missing settings and distinct actions, MCP entries, and search exclusions are recovered from the other copies. The open app receives the repaired projects and conversations together, and queued saves preserve recovered conversations until the app acknowledges them.

If duplicate projects have watches for the same PR, the repair keeps the watch associated with a recorded fix, including its settings and progress. If both watches record different fixes, both conversations survive and automation pauses until the user chooses a mode again.

Path comparisons cover Windows separators and drive roots, UNC paths, macOS casing, and WSL paths. A desktop's casing rules are not applied to a different remote machine.

# Testing

- Type checking and full lint pass.
- **414 tests across 21 targeted suites pass**, covering database upgrades, transcript and notes preservation, PR-watch collisions, desktop/remote/MCP creation, relocation conflicts, project settings, saved views, and live and queued draft handoff. The app's 34-test suite passed separately after hydration timeouts during parallel runs alongside Electron builds.
- An isolated Electron smoke run passed the selected app, settings, search, IPC, project, and composer checks with **0 console/runtime errors**. External provider, microphone, and remote-device integrations used deterministic mocks.
- All touched files pass formatting. The full format check still reports the existing `CLAUDE.md` issue.
- All 12 non-English catalogs have zero missing translations.
- Two independent reviews covered database/API and renderer/draft behavior. Their additional findings around notes, colliding PR watches, and queued drafts were fixed and covered by regression tests.
- GPT-6 Astra adversarially reviewed the settings repair and desktop reconciliation. Its findings were fixed, and its final check passed the earlier queued-save transcript-loss reproducer and deletion cases with no remaining blockers in that scope.
- The isolated test instance was stopped after verification; its artifacts are retained for inspection.

# Screenshots

Multiple projects remain available in the project switcher:

Multiple projects

Reopening the same folder shows an “already registered” notice and keeps one Kurdish-to-English project. This check supplied a fixture folder to the production registration action, with changed casing, separators, and a trailing slash; the project IDs stayed unchanged in memory and in SQLite:

Duplicate creation blocked


src/renderer/state/appStore.ts
```diff
@@ -56,15 +57,21 @@ export const useAppStore = create<AppStoreState>()(
             (persistedState as (Partial<AppStoreState> & { threads?: Thread[] }) | undefined) ??
             ({} as Partial<AppStoreState>);
 
+          const deduped = dedupeProjects(state.projects ?? currentState.projects);
```
SDSLeon: **Important — data loss** — deduping on hydrate makes the next persist cascade-delete the duplicate's thread transcripts.

**What's wrong:** `createDbStorage().getItem` remembers the **raw pre-merge** snapshot (`src/renderer/state/dbStorage.ts:65-67`), so after `merge` drops duplicate project `D`, the first write diffs against a snapshot that still contains `D` and emits `deletedProjectIds: ["D"]` (`dbStorage.ts:352`). `dbSyncChanges` runs `DELETE FROM projects WHERE id = ?` **before** any thread upsert (`src/main/db/sync.ts:45-50`), and `threads.project_id` is `REFERENCES projects(id) ON DELETE CASCADE` with `PRAGMA foreign_keys = ON` (`src/main/db/connection.ts:97`, `:62`). The cascade removes every thread row still holding `project_id = D`, which cascades again into `thread_runtime_items`, `thread_terminal_scrollback`, `thread_completed_turns`, `thread_context_usage` (`connection.ts:123-176`). The rehomed thread rows are re-upserted a few statements later, so the thread reappears with its title — but the runtime items are gone and nothing re-sends them. `deleteProjectNotes.run("D")` also drops the duplicate's notes. `dbSyncAll` has the identical ordering (`sync.ts:118-128`).

Concrete: DB has `P1` and `P2` both at `/repo`, thread `T` under `P2` with 200 runtime items. First launch after this change → `T` shows in the sidebar with an **empty transcript**, permanently. This fires on exactly the population the PR targets.

**Why guards don't stop it:** `T` is present in the outgoing snapshot, so it never enters `deletedThreadIds` — no `dbDiscardThreadRuntimeWrites`, no warning. The `isMainCreatedThreadUnmirrored` guard is explicitly bypassed when the project itself is going away (`sync.ts:145`).

**Fix:** make the merge non-destructive at the DB boundary. Either (a) reorder `dbSyncChanges`/`dbSyncAll` so project deletes run **after** the project and thread upserts, or (b) emit an explicit re-parent (`UPDATE threads SET project_id = ? WHERE project_id = ?`) before the delete. Consider moving the one-time repair into `migrate` behind `version: 6` instead of running it on every hydrate.

**Regression check:** DB-level test — seed `P1`, `P2` (same path), thread `T` under `P2` with 3 `thread_runtime_items`; run the hydrate + persist round trip; assert `T.project_id === P1` **and** `SELECT COUNT(*) FROM thread_runtime_items WHERE thread_id = T` is still 3.

src/renderer/app.tsx
```diff
@@ -451,10 +452,21 @@ const mainWindowCleanups: Array<() => void> = isMainWindow
       // Main-process project mutations must reach this whole-store snapshot
       // before its next dbSyncAll persistence write.
       readBridge().onProjectStateChanged(({ projects }) => {
-        useAppStore.setState({ projects });
+        const deduped = dedupeProjects(projects);
```
SDSLeon: **Important — dead UI state / lost drafts** — the broadcast dedupe path remaps threads only; view, draft panes, draft contents and experiments keep pointing at the dropped project id.

**What's wrong:** `merge` in `appStore.ts` calls `remapProjectView`, but this handler does not. If `view` is `{kind: "draft"|"experiment", projectId: <dropped>}`, `AppContent.tsx:297-299` finds no project and returns `<HomeView />` — the user's current screen vanishes mid-session with no explanation. `view.kind === "thread"` panes carry `draft:<projectId>#uuid` ids that neither path rewrites: `AppContent.tsx:327-338` collapses to Home when it's the only pane, or renders a permanently blank pane in a split. `draftContents[<dropped>]` and `composerDraftStorage("project", <dropped>)` are never moved, so a half-written prompt is silently unreachable — and the merge path makes this worse by remapping the view onto the canonical project whose composer is empty. Finally this handler passes the **deduped** id set to `reconcileExperiments` (`experimentStore.ts:110-120`), which hard-deletes every experiment whose `projectId` was the dropped id, losing candidate grouping and winner decisions.

**Why guards don't stop it:** `reconcileExperiments` is a reaper, not a remapper. `deleteProject` (`projectSlice.ts:136-200`) is the established template for retiring a project id — it clears composer storage, `draftContents`, follow-up queues, the view, and draft panes — and neither dedupe path does any of it.

**Fix:** extract one shared `applyProjectDuplicateRemap(state, duplicateIds)` used by **both** this handler and `merge`, covering `threads`, `view` (including `view.panes` draft ids via `parseDraftProjectId` + the draft-pane id builder), `draftContents`, `pendingComposerSeeds`, `pendingDraftWorktreeSelections`, and composer draft storage; add a `remapProjectExperiments(duplicateIds)` called **before** `reconcileExperiments`.

**Regression check:** fire the registered `onProjectStateChanged` listener with `[canonical, duplicate]` while `view = {kind:"draft", projectId: duplicate.id}` and a draft pane for the duplicate; assert the view and pane are re-keyed to the canonical id (not reset to Home), the draft text survives, and experiments are re-parented rather than deleted.

src/renderer/state/slices/projectSlice.ts
```diff
@@ -79,9 +80,13 @@ export interface ProjectSlice {
   reorderProjects: (sourceId: string, targetId: string, placement: ReorderPlacement) => void;
 }
 
-export const createProjectSlice: SliceCreator<ProjectSlice> = (set) => ({
+export const createProjectSlice: SliceCreator<ProjectSlice> = (set, get) => ({
   projects: [],
   addProject: (location, nameOverride, workspaceId) => {
+    const identity = projectIdentityKey({ location });
+    const existing = get().projects.find((project) => projectIdentityKey(project) === identity);
+    if (existing) return existing;
```
SDSLeon: **Important — silent no-op + config destruction** — the early return discards `nameOverride` and `workspaceId` and gives callers no way to tell reuse from creation.

**What's wrong:** `registerNewProject` (`src/renderer/actions/createProjectActions.ts:46-55`) calls `addProject(location, name, getActiveWorkspaceId())` and then unconditionally `autoDetectSetupScript(project)` + `captureProductEvent("project.added")`. Two concrete failures:

1. **Scripts wiped.** Project `Repo` at `/repo` has a hand-edited `setupScript` and 4 custom actions. User re-adds `/repo` (or `/repo/`, or a case variant). `autoDetectSetupScript` (`src/renderer/utils/gitHelpers.ts:138-148`) fires on the *existing* project and calls `updateProjectScripts(id, { setupScript, actions: [] })`, which replaces `scripts` wholesale (`projectSlice.ts:223-228`). All 4 actions and the custom script are gone — no prompt, no toast, no undo, persisted on the next sync.
2. **Silent no-op across workspaces.** `/repo` is filed in workspace A; the user is in workspace B and picks `/repo` via "use an existing folder". The returned project keeps `workspaceId: A`, so the sidebar (`workspaceSelectors.ts:36-38, 55-60`) shows nothing new, `openDraft` does not switch workspaces (`viewSlice.ts`), and the user lands on a draft for a project they cannot see or navigate back to. A `project.added` event is emitted for a project that was not added.

**Why guards don't stop it:** the folder picker never validates against the store and the modal has no duplicate affordance; the flow assumes `addProject` always yields a visible, fresh project. The `addProject` JSDoc still documents `workspaceId` as "workspace to file the new project into".

**Fix:** return creation status — `addProject` → `{ project, created: boolean }` (or add a sibling `findProjectByLocation`). In `registerNewProject`, gate `autoDetectSetupScript` and `captureProductEvent` on `created`; when `!created`, refile into (or switch to) the active workspace and show a localized "this folder is already a project" toast (12 catalogs per CLAUDE.md).

**Regression check:** add `/repo`, set custom `scripts`, add `/repo/` again → `scripts` unchanged and no setup detection; adding an already-registered folder from a different active workspace leaves it reachable and emits no `project.added`.

src/shared/projectIdentity.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export function projectLocationKey(location: ProjectLocation): string {
+  if (location.kind === "wsl") {
+    return `wsl:${location.distro.trim().toLocaleLowerCase()}:${normalizePath(location.linuxPath, false)}`;
+  }
+  return `${location.kind}:${normalizePath(location.path, location.kind === "windows")}`;
+}
+
+/** Include the owning desktop so equal paths on different machines remain distinct. */
+export function projectIdentityKey(project: Pick<Project, "location" | "remoteServerId">): string {
```
SDSLeon: **Important — cross-process divergence** — identity is enforced only in the renderer store, so the main process keeps creating duplicates that the renderer then deletes.

**What's wrong:** `register()` in `src/main/remote/projectCommands.ts:285-292` is the single funnel for `add-existing`, `create` and `clone`, and it inserts `makeProject(...)` with a fresh `randomUUID()` and no identity check. Its callers include the MCP tool `create_project` (`src/main/app-controls/mcp/tools/projects.ts:89-100`, returns `{created: true, project}`) and every paired mobile/remote client. Path: an agent calls `create_project` for `/repo` which already exists as `P1` → main commits `P2`, replies `created: true`, broadcasts → the renderer's `onProjectStateChanged` drops `P2`, the `projects` identity changes → `saveAppStore` → `appStoreRowChanges` puts `P2` in `deletedProjectIds` → main deletes the row it just wrote (with the cascade from the `appStore.ts:60` comment over any thread the caller already created in it). A retry loops the same create→delete cycle. `relocate` (`projectCommands.ts:245-267`) likewise rewrites `location` with no collision check and can make two existing projects collide.

**Why guards don't stop it:** `dedupeProjects` is renderer-local and writes nothing back to main, and a headless backend host (`src/backend/BackendDesktopServices.ts`) never runs it at all — there the duplicates simply persist, so `list_projects`/`get_project` and the desktop UI diverge permanently.

**Fix:** enforce identity on the authoritative side — in `register()` (and `relocate`), look up `deps.getProjects()` by `projectIdentityKey` and return the existing project instead of inserting a second row; keep the renderer dedupe as read-only repair only.

**Regression check:** `applyRemoteProjectCommand({kind:"add-existing", path})` twice with `C:\\repo\\` then `c:/REPO` returns the same project id and leaves one row; a renderer test asserting `onProjectStateChanged` with a duplicate queues no project deletion.

src/shared/projectIdentity.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function dedupeProjects(projects: readonly Project[]): {
+  projects: Project[];
+  duplicateIds: Map<string, string>;
+} {
+  const canonicalByIdentity = new Map<string, string>();
+  const duplicateIds = new Map<string, string>();
+  const unique: Project[] = [];
+  for (const project of projects) {
+    const existingId = canonicalByIdentity.get(projectIdentityKey(project));
```
SDSLeon: **Important — state corruption** — the synthetic Home project has a real location and can be deduped away (or swallow a real project).

**What's wrong:** `ensureHomeProject` (`src/renderer/state/slices/projectSlice.ts:100-113`) stores Home as a normal row with `id: HOME_PROJECT_ID`, `disabled: true`, and `location` = the user's home directory. `dedupeProjects` has no exemption for it. A user with a dotfiles project at `~` (or `C:\Users\me`) has two rows with the same identity key. `addProject` prepends new projects (`projectSlice.ts:94-96`) while Home is only prepended when missing, so the real project usually sorts first → **the Home row is dropped**, `duplicateIds` maps `__lightcode_home__ → <dotfiles id>`, and every Home thread is rewritten onto the dotfiles project. In the opposite ordering the real project is dropped and its threads are rehomed onto a `disabled: true` synthetic row.

**Why guards don't stop it:** `ensureHomeProject` matches only on `HOME_PROJECT_ID`, so it cannot detect that the id was remapped away — it just re-creates an empty Home. `isHomeScopeLocation` (`src/shared/homeScope.ts:22-27`) exists but is never consulted here. `HOME_PROJECT_ID`'s own comment states the goal is to "not … orphan existing Home threads"; this does exactly that, and combined with the cascade noted on `appStore.ts:60` the loss is irreversible.

**Fix:** skip identity bookkeeping for the synthetic row — `if (isHomeProjectId(project.id)) { unique.push(project); continue; }` — Home is never a duplicate and never a canonical target.

**Regression check:** `dedupeProjects([{id: "real", location: homeDir}, {id: HOME_PROJECT_ID, location: homeDir}])` keeps both rows and returns an empty `duplicateIds`, in both orderings.

src/shared/projectIdentity.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function normalizePath(path: string, caseInsensitive: boolean): string {
+  let normalized = path
+    .trim()
+    .replace(/\\/gu, "/")
+    .replace(/\/{2,}/gu, "/");
+  if (normalized.length > 1) normalized = normalized.replace(/\/+$/u, "");
+  if (caseInsensitive) normalized = normalized.toLocaleLowerCase();
+  return normalized;
```
SDSLeon: **Important — false merge on UNC + the fix is inert on macOS**

**What's wrong:**

1. **False merge (data loss).** `.replace(/\/{2,}/gu, "/")` collapses the leading `//` of a UNC path: `\\host\share\repo` → `/host/share/repo`, which is the same key as the drive-rooted `\host\share\repo`. `path.win32.isAbsolute` returns `true` for both, so `assertValidProjectPath` (`projectCommands.ts:83-90`) accepts either and MCP `create_project` can register both. Two genuinely different folders then collapse into one project — and per the `appStore.ts:60` comment that is destructive, not cosmetic. Same class: bare `"C:"` and `"C:\\"` both normalize to `c:`.
2. **The stated goal isn't met on macOS.** `kind` is chosen by host platform, not path shape (`src/shared/createProject.ts:22-32`: WSL UNC wins by shape, otherwise `win32 ? "windows" : "posix"`). Every macOS project is `posix`, and `projectLocationKey` compares posix paths case-**sensitively**. On the default case-insensitive APFS volume `/Users/me/Repo` and `/Users/me/repo` are one folder but stay two projects — the exact duplicate this PR exists to prevent, on the primary development platform. Linux is correct as written, so the discriminator must be the host filesystem, not the location kind.

**Why guards don't stop it:** the only tests here are a Windows separator/case case and a Windows dedupe case (`projectIdentity.test.ts`); neither UNC, nor bare-drive, nor posix case, nor the WSL branch is covered.

**Fix:** preserve a leading `//` when the input had exactly two leading separators, and keep `X:` distinct from `X:/`; pass an explicit, documented `caseInsensitive` capability derived from the host platform (darwin → true) into `projectLocationKey` rather than inferring it from `kind`.

**Regression check:** `\\host\share\repo` ≠ `\host\share\repo`; `C:` ≠ `C:\`; `/Users/me/Repo` === `/Users/me/repo` on darwin and ≠ on linux; a WSL distro/path case and a remote-vs-local separation case.

SDSLeon (changes requested):
## Deep review — Request changes

**Scope:** PR #764 @ `693a44c`, 7 files (5 source + 2 screenshots). Verdict: **6 Important — must fix before merge.**

### Must fix
- `src/renderer/state/appStore.ts:60` — deduping on hydrate emits a project deletion against the pre-merge snapshot; `dbSyncChanges` deletes the project row **before** upserting the rehomed threads, and `ON DELETE CASCADE` destroys those threads' runtime items. Threads survive with empty transcripts, irreversibly, for exactly the users this PR targets.
- `src/renderer/app.tsx:455` — the broadcast path remaps threads only: the current view, draft panes, draft contents and experiments keep pointing at the dropped id, so the user is bounced to Home, loses unsent draft text, and loses experiment grouping.
- `src/renderer/state/slices/projectSlice.ts:88` — the early return drops `nameOverride`/`workspaceId` and gives callers no create-vs-reuse signal: re-adding a folder wipes its custom setup script and actions, and re-adding from another workspace is a silent no-op.
- `src/shared/projectIdentity.ts:15` — identity is enforced only in the renderer; `projectCommands.register()` (MCP `create_project`, mobile/remote clients) still inserts duplicates that the renderer then deletes, so `create_project` returns an id that stops existing. Headless backend hosts never dedupe at all.
- `src/shared/projectIdentity.ts:27` — the synthetic Home row carries the home directory as its location and has no exemption, so a project at `~` deduplicates Home away (or gets swallowed by it), orphaning every Home thread.
- `src/shared/projectIdentity.ts:38-45` — `normalizePath` collapses the UNC `//` prefix (false merge of two distinct folders) and, because `kind` follows the host platform, leaves macOS posix paths case-sensitive — the duplicate this PR targets still occurs on macOS.

### Verified good
- The identity concept is right: location, not name, as the identity boundary, with `remoteServerId` included so equal paths on different hosts stay distinct — and `remoteProjection.ts:50-58` does set both `remoteServerId` fields synchronously, so remote projects do not collapse.
- Remote mirror rows are excluded from desktop persistence (`appStorePersistence.ts:14`), so the dedupe cannot delete a mirrored remote project.
- `merge` correctly remaps `view` via `remapProjectView` and rewrites `thread.projectId` — the hydration path's in-memory result is consistent (the gaps are at the DB boundary and in the sibling broadcast path).
- Targeted suites pass on the PR tree: `vitest run src/renderer/state/appStore.test.ts src/shared/projectIdentity.test.ts` → 94 passed; `tsc --noEmit` → 0 errors.
- `src/shared` is the right home for the module — it is consumed by both renderer and main, with no provider branching.

### Rejected after validation
- *Persist `version` not bumped to 6 for the dedupe repair*: `merge` already performs unconditional repair-on-read (`normalizeStoredThreadStatus`) and the on-disk shape is unchanged, and the PR adds a hydration regression test from the old shape. Moving the repair into `migrate` is still the better shape and is recommended in the `appStore.ts:60` comment, but it is not a standalone blocker.
- *`projectIdentityKey` reads `project.remoteServerId` rather than `location.remoteServerId`*: both fields are written together at every current write site; no live divergence exists.

SpookySandwich:
@SDSLeon, this follow-up is ready for another look.

Addressed all six findings in 0427bafc3ce02d478406c59e65203a5b1ae23fb4:

1. Duplicate repair now runs in database migration v42, before desktop or headless hydration. Full-state sync also moves dependent rows before deleting a duplicate. Regression tests preserve transcripts, notes, schedules, and PR watches.
2. Hydration and project broadcasts share the reference-remapping helpers. Views, split draft panes, saved layouts, experiments, unsent rich text, attachments, and queued draft insertions survive, including broadcasts that contain only the surviving project.
3. Creation reports whether it added or reused a project. Reuse preserves custom scripts/actions, applies an explicit name and workspace selection, skips setup detection and creation analytics, and shows a translated notice.
4. Main-process registration reuses the existing ID for desktop, remote/mobile, and MCP callers. Relocation rejects an already registered destination.
5. Home has a separate synthetic identity, so registering the home folder keeps both projects and their threads.
6. Path identity preserves UNC and drive-root distinctions and applies the host's macOS casing policy without applying it to remote Linux paths.

Two independent reviews also found collisions in notes, PR watches, and pending composer seeds. Those are fixed and covered by regressions. When two merged project records have different recorded PR fixes, both conversations remain and automation pauses until a mode is selected again.

Validation: 394 tests across 17 targeted suites pass, along with type checking and full lint. The final isolated Electron smoke run passed its automated scenarios and 10 mocked integration gates with zero console/runtime errors. All touched files pass formatting; the full formatter still reports the unchanged CLAUDE.md issue. The PR description now includes refreshed screenshots pinned to this commit. The test instance is stopped.

The new [CI run](https://github.com/Porabuild/Poracode/actions/runs/34736845720) is currently marked action_required with no jobs started.

src/main/db/projectDeduplication.ts
```diff
@@ -0,0 +1,158 @@
...
+export function repairDuplicateProjects(sqlite: SqliteDatabase): void {
+  const rows = sqlite
+    .prepare("SELECT * FROM projects ORDER BY sort_order ASC, rowid ASC")
+    .all() as ProjectRow[];
+  const { duplicateIds } = dedupeProjects(rows.map(rowToProject), {
+    caseInsensitivePosix: process.platform === "darwin",
+  });
+  rehomeProjectReferences(sqlite, duplicateIds);
+  const deleteProject = sqlite.prepare("DELETE FROM projects WHERE id = ?");
+  for (const duplicateId of duplicateIds.keys()) deleteProject.run(duplicateId);
```
SDSLeon, line 26: **Important — data loss** — migration keeps the newest duplicate and deletes the original project's settings.

**What's wrong:** `repairDuplicateProjects` chooses canonical as the first row in `ORDER BY sort_order ASC` (`src/main/db/projectDeduplication.ts:13-17`), then deletes every other row (`:19-20`) after rehoming threads/notes/watches. It never copies `scripts`, `mcp_servers`, `search_settings`, `gh_account`, `worktree_location`, `icon`, or `last_draft_config` onto the survivor (notes are merged; project-row fields are not).

`addProject` prepends (`projectSlice.ts:125`), and `dbSyncAll` writes `sort_order` from array index. A user who re-added `/repo` therefore has: newest nearly-empty duplicate at sort 0 (auto-detected setup script, `actions: []`) and the original configured project at sort ≥1 (custom setup script and actions, most threads). v42 makes the empty new row canonical and drops the original row. Threads survive on the new id; custom actions/MCP/search/git-account settings do not.

Concrete: `P_old` created 2024 with 4 actions + 200 threads, `P_new` created 2026 by re-adding the folder, sort_order 0. First launch after this change → chats reappear under `P_new` with an empty Actions list, permanently.

**Why guards don't stop it:** `mergeNotes` only covers `project_notes`. Tests seed the keeper at sort 0 (`migrations.test.ts:126-127`), which is the opposite of production prepend order.

**Fix:** within each identity group, pick canonical by earliest `created_at` (stable original), and before delete copy non-empty project settings from discarded rows onto canonical (same idea as `mergeNotes`). Do not use sidebar `sort_order` as the keeper.

**Regression check:** seed two projects at `/repo` — older row with custom `scripts.actions` and 3 runtime items, newer row with `sort_order = 0` and default/empty scripts; run v42; assert the surviving row still has the custom actions **and** the 3 runtime items.

SDSLeon (changes requested):
## Deep review — Request changes

**Scope:** PR #764 @ `6cec306` (fix commit `0427bafc` + merge of master), 55 files. Verdict: **1 Important — must fix before merge.**

### Must fix
- `src/main/db/projectDeduplication.ts:13` — v42 picks canonical by `sort_order` (newest-first from `addProject` prepend) and deletes the other project row without merging `scripts` / MCP / search / git-account fields. Upgrade of the exact duplicate population this PR targets drops the original project's custom settings while rehoming threads onto the empty re-add. See inline.

### Verified good
- Hydrate/sync cascade (`#1`): v42 + `rehomeProjectReferences` before `DELETE`, `dbSyncAll` rehomes before dropping missing project ids; `thread_runtime_items` asserted in `migrations.test.ts` and `sync.test.ts`.
- Live remaps (`#2`): `applyProjectStateSnapshot` rewrites view, draft panes, mounted composer contents, experiments (remap then reconcile), and panel contexts.
- Reuse (`#3`): `addProjectWithResult` returns `created`, applies name/workspace, gates `autoDetectSetupScript` / `project.added`, toasts the localized already-registered message.
- Authoritative identity (`#4`): `register()` reuses by `projectIdentityKey`; relocate 409s on conflict; MCP returns `created: false`.
- Home (`#5`): `projectIdentityKey` short-circuits `HOME_PROJECT_ID`; Home is never a duplicate or canonical target.
- Path keys (`#6`): UNC prefix preserved, `C:` ≠ `C:\`, macOS POSIX opt-in, remotes do not inherit the client's case policy, WSL distro fold + Linux path case.

### Rejected after validation
- Prior 6 Important findings on `693a44c` — all fixed on `0427bafc`.
- Optional remote `created` without a protocol bump — additive field; local MCP always sets it.

src/main/db/projectDeduplication.ts
SpookySandwich: Fixed in 16d835a5bf5d8c35bdb347747c789dff0afa7399. The repair now keeps the earliest-created project, regardless of sidebar order, and merges settings before deleting the other rows. The original project's non-empty choices win conflicts; missing settings, distinct actions/MCP entries, and search exclusions are recovered from the other copies. Explicit false values are preserved, and account/model/worktree configurations stay together.

The upgrade regression now uses production ordering: a newer default project at sort 0, an older project with four custom actions at sort 1, and three saved runtime items. It verifies the original ID, actions, recovered settings, and all three messages survive. Hydration and full-state sync use the same repair policy.

GPT-6 Astra's adversarial review also caught a desktop reconciliation gap and a queued-save race. Main now sends repaired projects and recovered conversations back to the desktop together, and protects recovered conversations until a later save acknowledges them. Its final check passed the original race reproducer and project-deletion cases, with no remaining blockers in that scope.

Validation: 414 targeted tests across 21 suites passed across the final runs; the app's 34-test suite passed separately after hydration timeouts while Electron was building. Type checking and full lint pass. The final isolated Electron smoke run passed its automated checks and five mocked integration gates with zero console/runtime errors, then shut down cleanly. All touched files pass formatting; the unchanged CLAUDE.md remains the only full-format warning. Ready for another look.

SDSLeon: Verified on 16d835a5b. `dedupeProjects` now keeps the earliest-created project per identity group and merges non-empty settings (scripts/actions, MCP servers, search settings, gh account, worktree location, icon, last draft config) before v42 deletes duplicate rows; upsert → rehome → delete ordering keeps runtime items intact. The regression test seeds production ordering (newer default row at sort 0, older configured row at sort 1, 3 runtime items) and asserts survivor id, actions, recovered settings, and all items. Touched suites (132 tests), typecheck, lint, and format pass locally. Resolving.

SDSLeon (approved):
Verified the final fix commit 16d835a5b: the v42 duplicate-project repair keeps the earliest-created project and merges settings before deleting duplicates, with regression coverage in production ordering. Touched suites, typecheck, lint, and format pass locally.

PR #775, opened 2026-09-15
feat(mcp): add workspace management and thread ungroup

SpookySandwich:
Adds MCP tools to list, create, and update workspaces, assign projects and Home threads to them, and ungroup threads. New projects and threads can inherit the caller's workspace.

Group changes protect experiment membership before updating rows or sending commands. The sidebar, MCP, and remote paths share one ungroup rule: remove the selected member and dissolve the group if only one member remains. Ungrouping preserves unrelated thread data.

Project registration preserves workspace assignment when reusing an existing folder.

Validation: 905 related tests passed (1 skipped), with focused group and HTTP regressions also passing. Typecheck, both lint passes, changed-file formatting, and localization checks pass. Repository-wide formatting flags only the unchanged Windows `CLAUDE.md` symlink placeholder.


src/main/app-controls/mcp/tools/threads.ts
```diff
@@ -471,7 +493,43 @@ export const threadTools: ToolDomain = {
...
+          "Pass group to assign a sidebar group, or ungroup/ungroupAll to remove one — not both.",
+        );
+      }
+      if (parsed.ungroup && parsed.ungroupAll) {
+        throw new Error(
+          "Pass ungroup to remove this thread from its group, or ungroupAll to dissolve the whole group — not both.",
+        );
+      }
+      if ((parsed.ungroup || parsed.ungroupAll) && !current.groupId) {
+        throw new Error(`Thread ${parsed.threadId} is not in a sidebar group.`);
```
SDSLeon: **Important — MCP ungroup can dissolve an experiment's candidate group and orphan the experiment**

**What is wrong.** Experiments are keyed by their sidebar group: `experimentActions.ts:153` creates every candidate with `groupId: experimentId`, and `SidebarThreadGroup.tsx` / `ThreadContextMenu.tsx` find the experiment via `experiments[thread.groupId]`. The desktop UI deliberately refuses to ungroup a candidate (`ThreadContextMenu.tsx:282` only offers "Remove from group" when `!isExperimentCandidate`). The new `ungroup` / `ungroupAll` paths here (and the mirrored ungroup in `threadCommands.ts` `set-group` and `remoteGroupCommandActions.ts`) have no experiment awareness.

**Trigger and impact.** `update_thread { threadId: <candidate>, ungroup: true }` on a 2-candidate experiment clears the target and, via the leftover-pair rule, the other candidate too; `ungroupAll` does the same for N candidates. `experimentStore.reconcileExperiments` only prunes by `projectId`, so the record survives with no thread carrying its `groupId`. The group header (the only route to `openExperiment` / `discardExperiment` besides the now-gated context menu entry) disappears, leaving a permanently orphaned experiment plus its candidate worktrees on disk with no UI path to clean up.

**Why existing guards do not prevent it.** The only guard is the renderer menu's `isExperimentCandidate` check; the MCP ingress and the remote `set-group`-without-`groupId` path bypass it.

**Smallest complete fix.** Reject at the boundary: after the "not in a sidebar group" check, throw when `current.groupId` is a persisted experiment id (a main-side reader already exists: `readPersistedExperiments()` in `src/main/remote/experimentOwnership.ts`; expose an `isExperimentGroup(groupId)` on the tool context). Apply the same rejection in `applyRemoteThreadCommand`'s ungroup branch so the HTTP route cannot reach it. Note the pre-existing `group` field already lets an agent re-home a candidate into another group; guarding both here would close that too.

**Regression check.** In `toolRegistry.test.ts`, seed a 2-thread group whose `groupId` equals a persisted experiment id, call `update_thread` with `ungroup: true` and `ungroupAll: true`, assert both throw, no `set-group` command is emitted, and no row is rewritten.

src/renderer/actions/remoteGroupCommandActions.ts
```diff
@@ -0,0 +1,48 @@
...
+    const target = state.threads.find((thread) => thread.id === threadId);
+    if (!target) return {};
+    if (groupId) {
+      return {
+        threads: state.threads.map((thread) =>
+          thread.id === threadId ? { ...thread, groupId, groupName: groupName ?? groupId } : thread,
+        ),
+      };
+    }
+    const previousGroupId = target.groupId;
```
SDSLeon, line 29: **Important — the "ungroup + dissolve leftover pair" rule is now implemented four times**

**What is wrong.** The same business rule (strip `groupId`/`groupName` from the target, then strip the last remaining member if exactly one is left, and collapse the `activeGroupId` view) exists in:
- `ThreadContextMenu.tsx:372-391` (pre-existing UI path)
- `src/renderer/actions/remoteGroupCommandActions.ts:9-48` (new, same process and same store as the menu)
- `src/main/remote/server/threadCommands.ts:207-227` (new)
- `src/main/app-controls/mcp/tools/threads.ts:571-596` (new)

**Impact.** Every future change to the rule must land in lockstep in all four places. The experiment guard requested in the other comment is a concrete example: it needs to be added to three of them. Two of the copies already live in the renderer against the same `useAppStore`.

**Smallest complete fix.** One pure helper, e.g. `dissolveGroupMembership(threads: readonly Thread[], threadId: string, options?: { all?: boolean }): { threads: Thread[]; clearedIds: string[] }` next to `Thread` in `src/shared/contracts` (or a sibling `threadGroups.ts`). Call it from the MCP handler (map `clearedIds` to `updateThreadRow` + `set-group` emits), from `threadCommands.ts`, and from this renderer action; have the `"ungroup"` case in `ThreadContextMenu.tsx` call `applyRemoteSetGroupCommand(thread.id, undefined, undefined)` so the renderer has a single copy.

**Behavior to preserve.** Dissolve only when exactly one member remains; the acting thread's removal is unconditional; view collapses to `{ kind: "thread", panes: [firstPane] }` only when `activeGroupId` was the affected group (matches today's menu behavior).

**Regression check.** Existing `remoteGroupCommandActions.test.ts` and `toolRegistry.test.ts` ungroup cases continue to pass; add one 3-member case asserting no dissolve on the shared helper.

SDSLeon (changes requested):
Review of #775 at `ddf9a4f6e7d00458b93ab3d63b5d2cfef6b8371b` (base `master` @ `0c99e7e`). Scope: the full diff (MCP tool handlers, workspace lookup/persistence, remote `set-group`/`set-workspace` contract, server thread commands, renderer mirror actions, app thread launcher, plugin metadata, tests), reviewed in full repository context.

**2 Important findings (must fix), both inline:**
1. `update_thread { ungroup | ungroupAll }` has no experiment guard, so it can dissolve an experiment's candidate group and orphan the experiment record and its worktrees. The desktop menu already blocks this; the MCP and remote paths bypass that guard.
2. The ungroup-and-dissolve-leftover rule is now implemented in four places (menu, renderer action, server command, MCP handler). The guard fix above would have to land in three of them, which is the compounding cost.

**Verified strengths.**
- Wire compatibility is sound: `set-group` `groupId` becoming optional and the new `set-workspace` kind are additive. No shipped wire sender (mobile client, provider-switch action) emits the new shapes; old senders still take the `groupId` branch. This matches the v9 additive-capability precedent, so leaving `PORACODE_REMOTE_PROTOCOL_VERSION` at 10 is acceptable.
- Persisted state needs no migration: `projects.workspace_id` and `threads.workspace_id` already exist and are mapped; `dbUpdateProject` writes `workspace_id`. Switching `ingressDeps.updateProject` from `dbUpsertProject` to `dbUpdateProject` also stops resetting sidebar sort order on rename.
- `update_project` cannot be clobbered by the renderer's full sync: `publishProjectsChanged` re-reads the DB after the write and the renderer replaces `projects` wholesale.
- `update_thread` validates every field before emitting; `workspaceId: null` survives `applyField`; `persistWorkspaces` uses the same guarded read/normalize/merge/write sequence as `update_settings` with no lost-update path against the renderer.
- MCP catalog mirrors (`BUILT_IN_MCP_SERVER_TOOL_NAMES`, plugin semver, SKILL.md) are all updated and covered by `mcpServer.test.ts`.
- Verification run in an isolated worktree at this head: targeted Vitest for `src/main/app-controls`, `src/main/remote`, `src/renderer/actions`, `src/shared/contracts`, `src/shared/remote` → 68 files / 868 tests passed; `pnpm run typecheck` clean.

**Important candidates checked and not posted.**
- Renderer ungroup collapsing a 3+-member group view to a single pane: this mirrors the existing `ThreadContextMenu` behavior exactly, so it is pre-existing parity, not a regression.
- Server-side `threadCommands.ts` ungroup/`set-workspace` have no test file: unreachable from any shipped client today, so not blocking.

Nit-level findings (a UI/MCP workspace-name length mismatch, the ungroup branch bypassing `applyField`, and three copies of the strip-group-fields helper) are omitted under this review's Important-only policy. This review does not assert CI status or merge safety beyond the checks listed.

SDSLeon (approved):
Validated the review fixes at 00f24fcda. Both previous Important findings are resolved: MCP and remote group changes guard persisted experiment ownership before writes/emits, including destination groups; ungrouping now uses the shared dissolveGroupMembership helper, with the context menu delegating to the renderer action. Reviewed preservation of unrelated row fields, pair/larger-group/all-member behavior, workspace registration integration, and backward-compatible command shapes. Regression coverage addresses the guarded paths and shared behavior. No Important findings remain; nits omitted. All exact-head GitHub CI checks passed, including typecheck, lint, format and eight test shards. Local test execution was attempted but dependency setup failed on missing electron/package.json, so no local test pass is claimed.

PR #777, opened 2026-09-19
fix(chat): render and open Home file citations

SpookySandwich:
Home threads showed raw `:codex-file-citation{...}` text instead of clickable file chips because formatting and file opening were grouped with project-only features.

Home now renders chips and opens the referenced files without enabling project browsing, indexing, or checkpoints. On mobile, Home hides workspace and folder browsing controls. Empty routes and failed reads no longer leave a blank screen, and explicit file links work at narrow and wide widths. Project threads retain their file browsing.

Absolute paths retain their identity, and switching file-editor contexts protects unsaved edits.

Validation: 149 focused regression tests, typecheck, both lint passes, changed-file formatting, and localization checks pass. App checks cover Home and project navigation, Back, real file reads, failed reads, and both layouts. Checks ran on Windows with emulated mobile widths; native PDF page rendering remains unverified.

Follow-up to #748.


src/mobile/views/WorkspaceView.tsx
```diff
@@ -50,7 +51,32 @@ export function WorkspaceView(props: {
   readonly onLaunchConflictResolverThread?:
     | ((input: ConflictResolverLaunchInput) => void)
     | undefined;
-}) {
+}
+
+export function WorkspaceView(props: WorkspaceViewProps) {
+  if (isHomeProjectId(props.filesTarget.project.id)) {
```
SDSLeon, line 57: **Important — Home thread's workspace chip now lands on an empty screen (both mobile layouts).**

**What is wrong:** this Home branch renders `FilesView` → `HomeFileView` unconditionally, but the existing entry point still reaches it without a file. `src/mobile/ThreadDetail.tsx:98` passes `onOpenWorkspace` for every thread, so `ThreadView` renders the `WorkspaceChip` for Home threads and `GitSummaryParts` fires `onOpen("files")` with no `file` search param. `HomeFileView` then has `initialFilePath === undefined`, `useOpenFile` never opens anything, and `MobileFileEditor` renders only the back chevron plus an empty path span (every body branch is `null`). The same happens for `?tab=changes` and `?folder=`, which this branch ignores. On the wide shell, `DesktopWorkspacePanel.show("files", id)` opens the panel with `showFilesTab={canBrowseProject}` false and `filesContent` `null` (lines 425/463), so the panel slides open empty.

**Impact:** on master the same tap opened the `$HOME` file tree. After this PR the only workspace entry point on a Home thread is a dead button that shows a blank pane, with no message. A failed Home read (`EACCES`, citation pointing at a directory) reaches the same blank state because `useOpenFile` sets `openFile` to `null`.

**Why guards don't catch it:** `canBrowseProject` gates the tool-rail button and panel content, not the chip or the route. `MobileFileEditor` was previously only mounted inside `openFile ? …`, so it never needed an empty state.

**Smallest complete fix:**
1. Do not pass `onOpenWorkspace` for Home threads in `ThreadDetail` (match the rail's `canBrowseProject` rule), and skip `show("files"/"git")` for Home in the wide shell.
2. Give `MobileFileEditor` a localized empty state when `openFile === null` (e.g. `<Trans>No file selected.</Trans>`), so the failed-read case is not blank either.

**Regression check:** extend `WorkspaceView.test.tsx` with a Home `filesTarget` and no `initialFilePath`, asserting a labeled empty state (or that the chip is absent from `ThreadView` for Home). Manually: open a Home thread on a narrow viewport, confirm no chip; visit `/workspace/<homeThreadId>?tab=files` and confirm a labeled state with a working Back; confirm a project thread still gets its tree on both shells.

SDSLeon (changes requested):
Review of #777 at `0fb5abb8c21acb457533b27bb3583e4e48045a1d` (reviewed against merge base `0c99e7e`).

**Scope:** renderer chat file actions (`chatPaneFileActions.ts`, `ChatPane.tsx`, `ItemMarkdownInner.tsx`, `parseProjectPathRef.ts`), editor-context switching (`fileEditorContext.ts`, `gitHelpers.ts`, `panelActions.ts`), editor overlay/LSP Home gating, and the mobile Home file surfaces (`WorkspaceView`, `FilesView`, `HomeFileView`, `MobileFileEditor`, `DesktopWorkspacePanel`, `useOpenFile`). Three independent review passes (desktop trust/lifecycle, mobile surfaces, rules/proof/duplication), then manual validation of every candidate.

**Important findings: 1** (inline below).

1. **Home thread workspace chip now opens an empty screen on mobile.** `ThreadDetail` still passes `onOpenWorkspace` for every thread, so the `WorkspaceChip` renders for Home threads and fires `onOpen("files")` with no `file` param. On the narrow shell the new Home branch in `WorkspaceView` renders `HomeFileView` with `initialFilePath === undefined`, so `MobileFileEditor` shows only a back chevron and an empty path span. On the wide shell `DesktopWorkspacePanel` opens the panel on the `files` tab while `showFilesTab={canBrowseProject}` is false and `filesContent` is `null`, so the panel slides open empty. On master the same tap opened the `$HOME` file tree. Smallest fix: gate the entry point (do not pass `onOpenWorkspace` for Home in `ThreadDetail`, and skip `show("files")` for Home in the wide shell), and give `MobileFileEditor` a localized empty state when `openFile === null` so a failed read (`EACCES`, directory target) is not also a blank page.

**Verified strengths:**
- The end-to-end Home path works as intended: absolute citation → `pathRefUrl` → `MdAnchor` → `resolveHomeFilePath` → `openFileInEditor` → `activateFileEditorContext` → `fileEditorStore.openFile` → `readExternalFile`. Relative Home refs resolve against the Home location; project-scope resolution (`normalizeChatProjectPath` + `resolveBareBasename`) is preserved verbatim.
- Forward-slash UNC output (`//host/share/...`) is still classified external by `isExternalPath`, `useOpenFile`, and `HomeFileView`, so no relativization against Home occurs.
- `resolveHomeFilePath` throw paths (`new URL`, `decodeURIComponent` on malformed sequences) run inside the async action and every caller attaches `.catch`, so no unhandled rejection.
- The new `window.confirm` in `activateFileEditorContext` is a net fix: all previous callers reset the root context silently and could drop dirty buffers. `remoteServerId` comparison is consistent with `buildFileEditorContext`.
- LSP cleanup still stops the old project server on a project→Home switch; only a Home id is skipped.
- Absolute Home opens are read-only on mobile, so Save never writes outside a project.
- No provider name or provider branch in shared files; no new `useMemo`/`useCallback`; no persisted or IPC shape changed. Lingui extract reports 0 missing in all 12 locales.

**Gates run on the frozen head:** vitest on the touched suites (96 files, 1082 tests passed, 1 skipped), `pnpm run typecheck` (pass after generating `@poracode/codex-protocol`), `pnpm run lint` (clean), `pnpm run fmt:check` (clean), `pnpm i18n:extract` (0 missing).

**Important candidates rejected:** `MdAnchor` now renders a file chip with no `onOpen` when the pane has actions but neither `projectLocation` nor `openProjectRelativePath` (project row not yet hydrated or deleted). The chip is inert rather than broken, and the state is transient, so this is a Nit, not blocking; mirroring the `MdCode` gate would tidy it. Nits are omitted from this review by policy. Not verified: native PDF rendering and physical mobile/WSL devices, as noted in the PR description.

SDSLeon (approved):
Validated the review fix at 6a0fe1f3 against the previously reviewed 0fb5abb8. The Home workspace/folder entry points are gated, explicit file citations remain available, empty Home routes and failed reads retain a labeled viewer with working Back navigation, and the wide shell no longer mounts an empty Home file panel. Project browsing remains available.

No Important findings in the follow-up after navigation/lifecycle and proof/reuse review passes. Verification: 114 focused tests across 11 suites passed; typecheck, both lint passes, changed-file formatting, and Lingui extraction passed (0 missing translations in all 12 non-English locales). No CI checks are reported by GitHub. Native PDF rendering and physical-device behavior were not revalidated. Nits omitted.

PR #780, opened 2026-09-21
feat(chat): typeset LaTeX in chat and markdown preview

SpookySandwich:
Typeset math in chat and Markdown previews with a shared `remark-math-extended` + KaTeX pipeline. Both `$…$` / `$$…$$` and `\(...\)` / `\[…\]` work, including replies that mix them. KaTeX runs after sanitization; malformed formulas stay local to the equation, and inline/fenced code stays literal.

Known parser tradeoffs: `$5 and $10` can become one formula, and `\[1\]` is display math. Use `\\[` for a literal backslash-bracket.

## Validation

- [x] **61 targeted tests**, including 22 math cases across chat and file preview: both syntaxes, mixed replies, incomplete delimiters, malformed formulas, code, escaping, and sanitization.
- [x] **Isolated Electron app:** mixed reply renders all eight equations; incremental runtime updates split inside delimiters finish correctly; the file-preview toggle renders the same equations; wide equations scroll within a 320px column. KaTeX fonts load and no renderer errors occur. These checks use fixture replies, without a provider API call.
- [x] Typecheck, touched-file lint/format, and production renderer build pass. The build reports CSS/chunk/sourcemap warnings.

## Before and after

Matching standalone previews of the same sample, with plain Markdown before and math rendering after.

| Syntax | Before | After |
| --- | --- | --- |
| Dollar | Dollar before | Dollar after |
| LaTeX | LaTeX before | LaTeX after |

<details>
<summary>Mixed-syntax reply in the isolated app and file preview</summary>

Chat:

Mixed math in the isolated app

File preview:

Mixed math in file preview

</details>


package.json
```diff
@@ -135,6 +135,7 @@
     "json5": "^2.2.3",
     "jsonc-parser": "3.3.1",
     "jsqr": "^1.4.0",
+    "katex": "^0.18.7",
```
SDSLeon, line 138: [P1] Match the KaTeX stylesheet to the renderer version

The direct `katex` dependency supplies `tailwind.css` with 0.18.7 CSS, but the lockfile resolves `rehype-katex` to KaTeX 0.16.47 for actual rendering. In 0.18 the stylesheet renamed classes such as `.base`, `.strut`, and `.sizing` to prefixed classes, so the 0.16 HTML loses its positioning and script-size rules in both chat and file preview. Rendering `\frac{1}{2}+x^2` with the installed renderer and applying these styles reproduces `.sizing` at 19.36px instead of the matching stylesheet's 13.552px, and `.base` becomes inline/static instead of inline-block/relative. The new tests only assert generated DOM and do not load this CSS. Align the direct KaTeX dependency with the renderer (e.g. 0.16.47), regenerate the lockfile, and verify fractions/superscripts with the actual bundled stylesheet.

SDSLeon (changes requested):
Reviewed all eight changed files at 833ac073f1651f715ec7258099bafadc81ba836f, including both rendering pipelines, sanitization/error handling, dependency resolution, styles, and tests. One Important finding: the KaTeX stylesheet does not match the renderer version and loses essential layout/sizing rules.

The shared pipeline preserves sanitization before generated markup and catches malformed equations locally. All 61 targeted tests passed (chatMathRendering, ItemMarkdownInner, ItemMarkdown, markdownPathRefRendering). A separate computed-style reproduction confirmed the version mismatch; the JSX tests do not load CSS. No unsafe trust bypass was found. GitHub reports no CI checks for this branch. Full Electron visual verification was not rerun. Nits omitted.

